PR #1, opened 2016-11-17
my resolution

zivoslav:
Here is my resolution.


src/Main.elm
```diff
@@ -1,16 +1,411 @@
...
+    Decode.succeed AudienceFolder
+        |: (field "id" Decode.int)
+        |: (field "name" Decode.string)
+        |: (field "parent" <| Decode.maybe Decode.int)
+
+
+audiencesDecoder : String -> Result String (List Audience)
+audiencesDecoder json =
+    json
+        |> Decode.decodeString (Decode.at [ "data" ] <| Decode.list audiencesDecoderItem)
```
turboMaCk, line 94: `Decode.decodeString (Decode.at [ "data" ] <| Decode.list` can be abstracted to own function `decodeCollection`

src/Main.elm
```diff
@@ -1,16 +1,411 @@
...
+
+
+audienceFoldersDecoder : String -> Result String (List AudienceFolder)
+audienceFoldersDecoder json =
+    json
+        |> Decode.decodeString (Decode.at [ "data" ] <| Decode.list audienceFolderDecodeItem)
+
+
+audienceFolderDecodeItem : Decoder AudienceFolder
+audienceFolderDecodeItem =
```
turboMaCk, line 84: `audienceFolderDecoder` is better fitting name I think

src/Main.elm
```diff
@@ -1,16 +1,411 @@
...
+
+
+audiencesDecoder : String -> Result String (List Audience)
+audiencesDecoder json =
+    json
+        |> Decode.decodeString (Decode.at [ "data" ] <| Decode.list audiencesDecoderItem)
+
+
+audiencesDecoderItem : Decoder Audience
+audiencesDecoderItem =
```
turboMaCk, line 98: inconsistent naming (`audienceFolderDecodeItem` vs `audiencesDecoderItem`). I think the best name is `audienceDecoder` in this case

src/Main.elm
```diff
@@ -1,16 +1,411 @@
...
+            Decode.succeed Authored
+
+        "shared" ->
+            Decode.succeed Shared
+
+        "curated" ->
+            Decode.succeed Curated
+
+        _ ->
+            Decode.fail (audienceType ++ " is not a recognized audienceType for Audience")
```
turboMaCk, line 122: Nice 👍

src/Main.elm
```diff
@@ -1,16 +1,411 @@
...
+        |: (field "name" Decode.string)
+        |: (field "type"
+                Decode.string
+                |> Decode.andThen audienceTypeDecoder
+           )
+        |: (field "folder" <| Decode.maybe Decode.int)
+
+
+audienceTypeDecoder : String -> Decoder AudienceType
+audienceTypeDecoder audienceType =
```
turboMaCk, line 110: yes... this is the right name

src/Main.elm
```diff
@@ -1,16 +1,411 @@
...
+
+
+update : Msg -> Model -> ( Model, Cmd Msg )
+update msg model =
+    let
+        _ =
+            Debug.log "UPDATE msg: " msg
+
+        _ =
+            Debug.log "UPDATE model: " model
```
turboMaCk, line 147: nice debugging strategy but probably should not be included in PR.

src/Main.elm
```diff
@@ -1,16 +1,411 @@
...
+            SelectRootDirectory folder ->
+                ( { model
+                    | currentPath = [ folder ]
+                    , selectedFolder = Just folder
+                  }
+                , Cmd.none
+                )
+
+            SelectBreadcrumbItem folder ->
+                ( { model | selectedFolder = Just folder }, Cmd.none )
```
turboMaCk, line 159: This looks bit weird. `currentPath` doesn't need update?

src/Main.elm
```diff
@@ -1,16 +1,411 @@
...
+    , currentPath : List AudienceFolder
+    , selectedFolder : Maybe AudienceFolder
+    }
+
+
+init : ( Model, Cmd Msg )
+init =
+    let
+        audienceFolder =
+            case Debug.log "audiencefolder" (audienceFoldersDecoder audienceFoldersJSON) of
```
turboMaCk, line 47: leftover `Debug.log`

src/Main.elm
```diff
@@ -1,16 +1,411 @@
...
+                Ok data ->
+                    data
+
+                -- @TODO: error reporting
+                Err err ->
+                    []
+
+        audiences =
+            case Debug.log "audiences" (audiencesDecoder Data.Audience.audiencesJSON) of
```
turboMaCk, line 56: same here

src/Main.elm
```diff
@@ -1,16 +1,411 @@
...
+                let
+                    newCurrentPath =
+                        setCurrentPath folder model.currentPath
+                in
+                    ( { model
+                        | currentPath = newCurrentPath
+                        , selectedFolder = Just folder
+                      }
+                    , Cmd.none
+                    )
```
turboMaCk, line 171: I'm still not sure if this is how it should work. I think one action/Msg should be enough just by looking at parameters.

src/Main.elm
```diff
@@ -1,16 +1,411 @@
...
+        ]
+
+
+viewRootFolderNavigationItem : Maybe AudienceFolder -> AudienceFolder -> Maybe (Html Msg)
+viewRootFolderNavigationItem selectedFolder folder =
+    let
+        -- @TODO: show state open/close of root folder is not correct for selected subfolders
+        isOpened =
+            case selectedFolder of
+                Just selectedFolder_unwrap ->
```
turboMaCk, line 242: using `_` in variable name is inconsistent.

src/Main.elm
```diff
@@ -1,16 +1,411 @@
...
+        Nothing ->
+            Nothing
+
+
+
+-- UTILS
+
+
+isOpenFolder : AudienceFolder -> AudienceFolder -> Bool
+isOpenFolder selectedFolder folder =
```
turboMaCk, line 388: api of this function seems a bit confusing.

src/Main.elm
```diff
@@ -1,16 +1,411 @@
...
+    if folder.id == selectedFolder.id then
+        True
+    else
+        False
```
turboMaCk, line 392: there is quite some redundant code.

```
folder.id == selectedFolder.id
```

will do the same trick (since it's expression which return Boolean)

turboMaCk (changes requested):
Please ignore Request for changes.. this is only to prevent merges. Thanks!

src/Main.elm
```diff
@@ -1,16 +1,411 @@
...
+    ul [ id "breadcrumbPanel", class "menu-list" ] <|
+        List.map
+            (viewBreadcrumbItem selectedFolder)
+            (List.reverse currentPath)
+
+
+viewBreadcrumbItem : Maybe AudienceFolder -> AudienceFolder -> Html Msg
+viewBreadcrumbItem selectedFolder folder =
+    let
+        -- @TODO: Im considering, that is better pattern matching of selectedFolder:Maybe AudienceFolder in "isOpenFolder" function???
```
turboMaCk, line 286: 👍 Yeah... Maybe you're right... Anyway even this way do not seems too bad to me

PR #2, opened 2016-12-28
t/first-step@diegonc Hierarchical browser widget

diegonc:
Hello

First of all, thank you very much for taking the time to publish this challenge. It got me an excuse to dive into and work with Json decoders; which is something I've been delaying.

The boilerplate is in the `dist` directory. It includes an index HTML and Bootstrap's CSS and glyphicons font. The latter quickly got me somewhat pretty and allowed to do a minimal layout for the debugging info.

When implementing the Audience decoder I had to assume the `user` JSON type corresponded to the `Authored` Elm type; in the comments there's some data analysis I've done. [`Data.Json.*`]

I included a module for API related functions which should return Tasks or Cmds in the future. Currently they are just mocks and work synchronously. [`Data.Api.Endpoints`]

The main data structure is located in the `Hierarchy` module. I have to admit I'd prefer to use a tree and its zipper but the current table based implementation was more appropriate to my FP knowledge and allowed me to get something working instead of nothing.

It is composed of:

* A nodes database (both AudienceFolder and Audience) indexed by ID.
* An index that given an ID provides its children.
* A list of top-level nodes.
* A breadcrumb indicating the path from the root of the tree to the currently displayed node.

Some things I'd like to improve/reconsider are:

* Navigation state (a.k.a. breadcrumb) could be separated from this data structure. But its need to access the children index and node database are good reasons to simplify and locate them in the same module.
* The list of top-level nodes could very well be a `Set` instead; as well as the values in the `Dict` where the children index is stored.
* The current API tends to be a bit verbose with respect to error handling.
* The model has two states (error & loaded) which forces to pattern match in update twice; there has to be some better way.
* In the view, the presentation of errors could be improved.

If you are still reviewing PR, it would be great to receive some comments about this one. (you said 100% ok 😝 )

All in all, I had fun doing this challenge; implementing `buildInitialHierarchy` even provided a little secondary mission I got help with in Slack documented [here](https://gist.github.com/diegonc/79199c5c97315d54efa3a63fa6b6ad0d). I want to thank you again for publishing it.


src/Data/Json/Audience.elm
```diff
@@ -0,0 +1,61 @@
+module Data.Json.Audience exposing (audienceDecoder)
+
+import Data.Json.Utils exposing (ref)
+import Json.Decode as JD exposing (andThen, (:=))
+import Data.Audience exposing (..)
+
+
+{-|
+
+Doing some data exploration with lodash (sorry for the heresy :-)
```
turboMaCk, line 10: No problem at all. Handy way to quickly explore data set like this since you can use various JS consoles to be effective :+1:

src/Data/Json/Audience.elm
```diff
@@ -0,0 +1,61 @@
...
+
+    {"type":"curated","curated":true,"shared":false} : 156
+    {"type":"curated","curated":true,"shared":null}  : 7
+    {"type":"shared","curated":false,"shared":true}  : 10
+    {"type":"shared","curated":null,"shared":true}   : 30
+    {"type":"user","curated":null,"shared":false}    : 9
+
+I'll assume the type field is enough to distinguish the audience type where
+the value "user" translates to Authored, "curated" to Curated and "shared"
+to Shared.
```
turboMaCk, line 31: And you're right (or at least I think.. if not then we have bug in production:D) 👍

src/Data/Api/Endpoints.elm
```diff
@@ -0,0 +1,21 @@
...
+import Data.Audience exposing (Audience, audiencesJSON)
+import Data.AudienceFolder exposing (AudienceFolder, audienceFoldersJSON)
+import Json.Decode as JD exposing ((:=))
+
+
+{-| GET /api/audiences
+-}
+audiences : Result String (List Audience)
+audiences =
+    JD.decodeString ("data" := (JD.list audienceDecoder)) audiencesJSON
```
turboMaCk, line 14: Not to say this is bad but I've wrote this abstraction (it's elm 0.18 but was 0.17 originally):

```elm
decodeCollection : Decoder a -> Decoder (List a)
decodeCollection =
    list >> Decode.at [ "data" ]
```

can be handy.

src/Data/Json/Audience.elm
```diff
@@ -0,0 +1,61 @@
...
+
+{-| Json decoder for an Audience value
+-}
+audienceDecoder : JD.Decoder Audience
+audienceDecoder =
+    JD.object4 Audience
+        ("id" := JD.int)
+        ("name" := JD.string)
+        ("type" := audienceTypeDecoder)
+        ("folder" := ref)
```
turboMaCk, line 61: I recommend using Json.Decode.Extra and applicative style. Looks like:

```elm
categoryDecoder : Decoder Category
categoryDecoder =
    succeed Category
        |: field "id" int
        |: field "name" string
        |: field "parent_id" (maybe int)
```

more info: http://www.troikatech.com/blog/2015/08/17/decoding-larger-json-objects-in-elm/ https://robots.thoughtbot.com/decoding-json-structures-with-elm and many more...

src/Hierarchy.elm
```diff
@@ -0,0 +1,267 @@
...
+
+
+{-|
+A container that can hold either an Audience or AudienceFolder.
+
+Folders are paired with its children count. (TODO only direct or all of them?)
+-}
+type Node
+    = Folder Int AudienceFolder
+    | File Audience
```
turboMaCk, line 27: Nice Type! 👍

src/Hierarchy.elm
```diff
@@ -0,0 +1,267 @@
...
+        , childIndex = Dict.empty
+        , nodeDB = Dict.empty
+        }
+
+
+{-|
+Navigates a given amount of levels up in the hierarchy.
+-}
+goup : Int -> Hierarchy -> Hierarchy
+goup levels (Hierarchy h) =
```
turboMaCk, line 61: you can ad `as hierarchy`...

src/Hierarchy.elm
```diff
@@ -0,0 +1,267 @@
...
+    case h.breadcrumb of
+        [] ->
+            Hierarchy h
+
+        _ :: rest ->
+            if levels <= 0 then
+                Hierarchy h
```
turboMaCk, line 68: ... and return old value. But is just matter of taste in this case (performance overhead is minimal).

src/Hierarchy.elm
```diff
@@ -0,0 +1,267 @@
...
+                    (Result.fromMaybe "Currupted child index: contains inexistant nodes")
+                |> Maybe.withDefault (Err "Corrupted breadcrumb: current node does not exist")
+
+
+{-| Add a folder to the hierarchy
+-}
+registerFolder : AudienceFolder -> Hierarchy -> Result String Hierarchy
+registerFolder folder (Hierarchy h) =
+    if Dict.member folder.id h.nodeDB then
+        Err <| "Node with ID " ++ (toString folder.id) ++ " has already been added"
```
turboMaCk, line 145: alternatively you can update old member. I can imagine scenario where it will be useful. Anyway I 👍 with not doing implicit magic if it's not required.

src/Hierarchy.elm
```diff
@@ -0,0 +1,267 @@
...
+
+{-| Adds a node ID to the list of node references
+This is a helper function used in a `Dict.update` call applied to the
+children index.
+-}
+addChild : Int -> Maybe (List Int) -> Maybe (List Int)
+addChild id maybeList =
+    maybeList
+        |> Maybe.map ((::) id)
+        |> Maybe.orElseLazy (\_ -> Just [ id ])
```
turboMaCk, line 227: There it will be much simplier to have `List a` instead of `Maybe (List a)` since `Nothing` and `Just []` are kind of similar in this case. Generally I think using `Maybe (List a)` makes sense only in applicative monadic or similar operations.

src/Hierarchy.elm
```diff
@@ -0,0 +1,267 @@
...
+            compare x.name y.name
+
+        ( Folder _ x, Folder _ y ) ->
+            compare x.name y.name
+
+        ( File _, Folder _ _ ) ->
+            GT
+
+        ( Folder _ _, File _ ) ->
+            LT
```
turboMaCk, line 267: This solution is not bad but I think it show that it's not that trivial and sometimes choosing simple solution can quite explode in a way you start handling cases all over the place. Personally I would abstract from the thing little bit, solve just that abstract problem and then reused this abstraction. I've built small package for solving hierarchy navigation abstractly if you're interested. It's what we are using: https://github.com/turboMaCk/chae-tree

It's really nice that you separated this from `Model, Update, View` logic but I would push it one level of abstraction up.

Hard to say what solution is better. Personally I like to break problem into smaller ones even though as a whole it can lead to slightly bigger solution it's bit more maintainable in my experience since business logic is not missed up with algorithms.

Anyway I'm giving you 👍 👍 👍. Good job in dealing with this. This implementation is indeed one of the best I saw for similar problems.

src/Main.elm
```diff
@@ -1,16 +1,265 @@
...
+                    (\a b -> b `Result.andThen` (H.registerFile a))
+                    withFolders
+                    audiences
+    in
+        folders `Result.andThen` \a -> audiences `Result.andThen` \b -> buildHierarchy a b
+
+
+type Model
+    = Error String
+    | Loaded (Maybe H.Node) Hierarchy
```
turboMaCk, line 42: again you're really good in finding expressive types 👍

src/Main.elm
```diff
@@ -1,16 +1,265 @@
...
+
+        audiences =
+            Api.audiences
+
+        buildHierarchy folders audiences =
+            let
+                withFolders =
+                    List.foldl
+                        (\a b -> b `Result.andThen` (H.registerFolder a))
+                        (Ok H.empty)
```
turboMaCk, line 29: I would prefer `<|` instead of `()` in this case but you know... just syntax 💩

src/Main.elm
```diff
@@ -1,16 +1,265 @@
...
+    Result.unpack Error (Loaded Nothing) buildInitialHierarchy
+
+
+
+-- UPDATE
+
+
+type Msg
+    = GoDown Int
+    | GoUp Int
```
turboMaCk, line 56: here it starts to be a bit confusing...
- GoDown Int -> Int is id, right?
- GoUp Int -> Int is number of levels up right?

Quick solution can be to add type alias to Int like `type alias Id = Int` and use it in case of `GoDown`. Not bullet proof but better I think.

src/Main.elm
```diff
@@ -1,16 +1,265 @@
...
 -}
-main : Html msg
+main : Program Never
 main =
-    Html.text "There will be app soon!"
+    Html.beginnerProgram
+        { model = init
+        , view = view
+        , update = update
+        }
```
turboMaCk, line 265: I really appreciate that you managed to keep `Main` so simple and easy to understand. Giving you few more of these 👍 👍 👍  again.

turboMaCk (commented):
Hi @diegonc!

Thank for your implementation and really great description.

I'm sorry for late response but I came around this PR during Christmas holidays and thought I would replay in beginning of January but I've forgot:(

I'm actually amazed with what you've done. You really showed that your ability to thing about code and problem from beginning to end is amazing. It was really pleasure to read through this especially when I take to account both description and code together.

I've tried to comment implementation as much as possible with all my thoughts. But honestly most of the stuff is just meter of taste.

One thing that you didn't mention though is if you're interested in job or not. Please let me know in case you would consider this opportunity. Otherwise I wish you the best of luck not just in work.

Cheers!
Marek

src/Data/Json/Audience.elm
diegonc: Interesting, that looks nice in terms of extensibility.

src/Hierarchy.elm
diegonc: Awesome tip!

src/Hierarchy.elm
diegonc: The Maybe comes from `Dict.update` because the entry may not be there yet. But yes, simplifying is good.

src/Hierarchy.elm
diegonc: Nice! I'll have a look.

I had to solve a similar problem at work a few months ago, i.e. building a tree from some data, where the order of the nodes is not ideal; some child node could be missing some of its parents in the partial tree.

And I ended up with a list of orphans which is checked when new nodes arrive; actually one list per-level, meh, it only works for known and bounded tree depths when you can tell what level each node goes in. 😛

Maybe I can find some inspiration in your library.

src/Main.elm
diegonc: Right, it's quite hard tell what each Int means as it is expressed. An Id type would help, indeed.

diegonc:
Thanks for the review! Nice and sensible advice in it 👍 

About the position, I didn't mention anything because:

* I thought it would have been already taken :)
* Relocation is not something I can consider; college mainly
* Remote means I'm 3 hours early; I'm afraid it could result a bit uncomfortable
* Speaking English is something I need to get better at

PR #4, opened 2017-05-18
Thanks for such a great challenge

kraklin:
This was really fun. I have finally an excuse to try JSON decoders and even try to implement my own tree structure in Elm. I learn a few tricks studying another tree implementations and decode my first JSON in Elm  😃 

I didn't bother with nice GUI, because I wanted to keep things simple and runnable with elm reactor, so there are no CSS or other features, just plain HTML and lists.

Anyway - Thanks 👍


turboMaCk:
hey @kraklin I'm glad you've enjoyed hacking on this!

I'll review your code later to give you some feedback. We don't really care about UI look that much in case of challenges like this so don't worry. For now just:

kraklin:
Sure, take your time. I'm looking forward to your ideas and suggestions ;)

Janiczek:
@kraklin a piece of history, isn't it :smile:

kraklin:
Yup, I have been cleaning of stale branches :) But yeah, nostalgic as hell :D and it has been "only" 3 years

PR #6, opened 2018-04-15
Roman Frołow - Challenge

rofrol:
Thanks for the challenge. I needed to learn more deeply json decoders.

Specifically for such big json, I didn't wanted to overlook some property. That's why I haven't use `andMap` but custom decoder using `keyValuePairs`.

So bonus features are:
- [x] Strict decoding of keys with warnings in console
- [x] Natural ordering
- [x] Ellipsis for text when it cannot be broken

output

### DIFF
--- a/src/Data/Audience.elm
+++ b/src/Data/Audience.elm
@@ -1,12 +1,21 @@
-module Data.Audience exposing (AudienceType(..), Audience, audiencesJSON)
+module Data.Audience exposing (Audience, AudienceType(..), audiencesDecoder, audiencesJSON)
 
 {-| Data.Audiences module
 
 This module implements everything related to audience resource.
 
+
 # Interface
+
 @docs AudienceType, Audience, audienceJSON
+
 -}
+
+import Data.DecoderUtils exposing (fromResultRecord, partialDecodeField)
+import Data.Expression exposing (Expression, defaultExpression, expressionDecoder)
+import Json.Decode exposing (..)
+import Json.Decode.Extra exposing ((|:))
+
 
 -- Type definition
 
@@ -24,9 +33,95 @@ type AudienceType
 type alias Audience =
     { id : Int
     , name : String
+    , expression : Expression
     , type_ : AudienceType
+    , category : Maybe String
     , folder : Maybe Int
     }
+
+
+audienceDecoder : Decoder Audience
+audienceDecoder =
+    succeed
+        (\audienceRecord ->
+            let
+                { id, name, expression, curated, shared, category, folder } =
+                    audienceRecord
+
+                audienceType =
+                    if curated == Just True then
+                        Curated
+                    else if shared == Just True then
+                        Shared
+                    else
+                        Authored
+            in
+            Audience id name expression audienceType category folder
+        )
+        |: audienceRecordDecoder
+
+
+type alias AudienceRecord =
+    { id : Int
+    , name : String
+    , expression : Expression
+    , type_ : String
+    , curated : Maybe Bool
+    , shared : Maybe Bool
+    , category : Maybe String
+    , folder : Maybe Int
+    }
+
+
+defaultAudienceRecord : AudienceRecord
+defaultAudienceRecord =
+    AudienceRecord -1 "" defaultExpression "" Nothing Nothing Nothing Nothing
+
+
+feedNameL : (String -> AudienceRecord -> Decoder AudienceRecord) -> List ( String, Value ) -> AudienceRecord
+feedNameL decoder elements =
+    List.foldl (fromResultRecord decoder) defaultAudienceRecord elements
+
+
+audienceRecordDecoder : Decoder AudienceRecord
+audienceRecordDecoder =
+    map (feedNameL decodeField) partialDecodeField
+
+
+decodeField : String -> AudienceRecord -> Decoder AudienceRecord
+decodeField name record =
+    case name of
+        "id" ->
+            succeed (\value -> { record | id = value }) |: int
+
+        "name" ->
+            succeed (\value -> { record | name = value }) |: string
+
+        "expression" ->
+            succeed (\value -> { record | expression = value }) |: expressionDecoder
+
+        "type" ->
+            succeed (\value -> { record | type_ = value }) |: string
+
+        "curated" ->
+            succeed (\value -> { record | curated = value }) |: nullable bool
+
+        "shared" ->
+            succeed (\value -> { record | shared = value }) |: nullable bool
+
+        "category" ->
+            succeed (\value -> { record | category = value }) |: maybe string
+
+        "folder" ->
+            succeed (\value -> { record | folder = value }) |: maybe int
+
+        _ ->
+            fail <| "Unhandled field"
+
+
+audiencesDecoder : Decoder (List Audience)
+audiencesDecoder =
+    field "data" (list audienceDecoder)
 
 
 
@@ -43,6 +138,7 @@ This is how we usually deal with making non-breaking continuous changes
 from old version of API to new one.
 
 You're free to use any strategy to decode JSON.
+
 -}
 audiencesJSON : String
 audiencesJSON =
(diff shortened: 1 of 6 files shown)


src/Data/Audience.elm
```diff
@@ -24,11 +33,97 @@ type AudienceType
...
+            succeed (\value -> { record | shared = value }) |: nullable bool
+
+        "category" ->
+            succeed (\value -> { record | category = value }) |: maybe string
+
+        "folder" ->
+            succeed (\value -> { record | folder = value }) |: maybe int
+
+        _ ->
+            fail <| "Unhandled field"
```
turboMaCk, line 119: redundant `<|`

src/Data/DecoderUtils.elm
```diff
@@ -0,0 +1,18 @@
...
+
+
+fromResultRecord : (String -> b -> Decoder b) -> ( String, Value ) -> b -> b
+fromResultRecord decoder ( name, value ) record =
+    case decodeValue (decoder name record) value of
+        Ok newRecord ->
+            newRecord
+
+        Err error ->
+            Debug.log ("[Warning]: " ++ error ++ " for name = " ++ name ++ ", value = " ++ toString value) record
```
turboMaCk, line 18: interesting idea.

src/Data/Expression.elm
```diff
@@ -0,0 +1,350 @@
...
+                "question": "q311"
+            }
+        ]
+    }
+"""
+
+
+type Expression
+    = RootOr (List BranchOrLeaf)
+    | RootAnd (List BranchOrLeaf)
```
turboMaCk, line 212: What's the purpose of this? Just ensuring there can't be a leaf without node in expression?

src/Data/Expression.elm
```diff
@@ -0,0 +1,350 @@
...
+    {
+        "options": [
+            "q25_6"
+        ],
+        "question": "q25"
+    }
+"""
+
+
+type alias OptionsQuestionNot =
```
turboMaCk, line 269: why this name?

src/Data/Expression.elm
```diff
@@ -0,0 +1,350 @@
...
+    { options : List String
+    , question : String
+    , not : Maybe Bool
+    , minCount : Maybe Int
```
turboMaCk, line 273: pro tip -> null in here can be defaulted to `1` but you can't probably tell without knowing the specification of audiences that's not part of this.

src/Data/Expression.elm
```diff
@@ -0,0 +1,350 @@
...
+        ],
+        "question": "q25"
+    }
+"""
+
+
+type alias OptionsQuestionNot =
+    { options : List String
+    , question : String
+    , not : Maybe Bool
```
turboMaCk, line 272: this should be `Bool`. if there are any broken data then it should probably be false by default.

src/Data/Audience.elm
```diff
@@ -24,11 +33,97 @@ type AudienceType
...
+        |: audienceRecordDecoder
+
+
+type alias AudienceRecord =
+    { id : Int
+    , name : String
+    , expression : Expression
+    , type_ : String
+    , curated : Maybe Bool
+    , shared : Maybe Bool
```
turboMaCk, line 70: maybes here are strange :/

src/Data/Expression.elm
```diff
@@ -0,0 +1,350 @@
...
+
+defaultOptionsQuestionNot : OptionsQuestionNot
+defaultOptionsQuestionNot =
+    OptionsQuestionNot [] "" Nothing Nothing []
+
+
+type BranchOrLeaf
+    = Item OptionsQuestionNot
+    | Or (List BranchOrLeaf)
+    | And (List BranchOrLeaf)
```
turboMaCk, line 286: Personally, I would use slightly different data type to express this AST. Say something like:

```elm
-- question+options leaf
type alias LeafRecord = { questionId : String, not : Bool ..... } 

-- Supported logic operations
type LogicOperator = And | Not

-- general recursive expression
type Expression = Node LogicOperator (List Expression) | Leaf LeafRecord

-- ensure there is at least one node at root.
type ExpressionRoot = Root LogicOperator (List Expression)
````

File: src/Main.elm [hunk 9]
```diff
@@ -1,16 +1,219 @@
...
+            [ color (hex "#ecf2f6"), backgroundColor (hex "#346f97"), cursor default ]
+        , Css.Foreign.class FolderCountClass
+            [ borderRadius (px 4), color (hex "#779aac"), backgroundColor (hex "#295a78"), padding (px 4), marginLeft (px 6) ]
+        , Css.Foreign.class AudienceEditIconClass
+            [ marginLeft auto, minWidth (px 30), displayFlex, justifyContent flexEnd ]
+        ]
+
+
+type alias Model =
+    { breadcrumbs : List Int
```
turboMaCk, line 70: do you really need list in here? I think you can just store `Maybe Int` of the current folder. Ancestors are then recursive function. There would some performance penalty for those lookups but it will limit state this component manages so less potential bugs and state synchronizations when structure starts to be editable.

src/Main.elm
```diff
@@ -1,16 +1,219 @@
...
+
+        audienceFolders =
+            decodeString audienceFoldersDecoder audienceFoldersJSON
+                |> Result.withDefault []
+
+        currentFolderList =
+            audienceFolders
+                |> List.filter (\item -> Just item.id == currentLevelMaybe)
+                |> List.head
+                |> Maybe.andThen (\item -> Just [ folderHtml True item ])
```
turboMaCk, line 108: :+1: glad I see `andThen` over pattern matching on this:)

src/Main.elm
```diff
@@ -1,16 +1,219 @@
...
+                                [ fromUnstyled <| icon edit ]
+                            ]
+                    )
+
+        audiencesInFoldersCounts =
+            List.foldl
+                (\element accumulator ->
+                    case element.folder of
+                        Just parentId ->
+                            Dict.insert parentId (Dict.get parentId accumulator |> Maybe.withDefault 0 |> (+) 1) accumulator
```
turboMaCk, line 138: Question: Does usage of dictionary here affects order or not?

src/Main.elm
```diff
@@ -1,16 +1,219 @@
...
+                        Nothing ->
+                            accumulator
+                )
+                audiencesInFoldersCounts
+                audienceFolders
+
+        currentLevelAudienceFolders =
+            audienceFolders
+                |> List.filter (\item -> item.parent == currentLevelMaybe)
+                |> List.sortWith (NaturalOrdering.compareOn .name)
```
turboMaCk, line 162: oh now I see. you're reordering things anyway here.

turboMaCk (approved):
Nice PR. It's clear you have quite some experience with Elm, this is definitely professional level implementation :+1: I even learned some things myself from this (elm css and natural ordering are things I'm not much familiar with).

I have some minar points and suggestions to some tiny details but parser for audience expression was kind of out of scope of this anyway so that leaves maybe just a decision to store all breadcrumbs list which can be potentially problematic if you other features are added to this but is a valid decision in scope of just this implementation.

well done!

src/Data/Expression.elm
rofrol: Exactly.

src/Data/Expression.elm
rofrol: Naming is a thing to improve. I just wanted names to navigate quickly when prototyping.

src/Data/Expression.elm
rofrol: Ok.

src/Data/Expression.elm
turboMaCk: makes sense.

src/Data/Expression.elm
rofrol: Make sense.

src/Data/Audience.elm
rofrol: It was meant to capture all possible variables to AudienceRecord and make decision afterwards how to create Audience. It was faster for me to make it like that.

src/Main.elm
rofrol: No need. I just wanted to experiment how would I implement breadcrumbs in Elm.

src/Data/Expression.elm
rofrol: What's the definition of `Root` and `Node`?

PR #7, opened 2018-05-21
Implement challenge

caitt:
- use json decoders to parse JSON
- use tree to encode data

### DIFF
--- a/src/Main.elm
+++ b/src/Main.elm
@@ -1,16 +1,96 @@
 module Main exposing (main)
 
-import Html exposing (Html)
+import Data.Audience exposing (Audience, audiencesDecoder)
+import Data.AudienceFolder exposing (AudienceFolder, foldersDecoder)
+import Html exposing (..)
+import Html.Attributes exposing (class)
+import Html.Events exposing (onClick)
+import Json.Decode exposing (Decoder)
+import Json.Decode.Pipeline exposing (..)
+import Tree exposing (..)
 
 
--- Import Modules
+type alias Model =
+    ( Tree, List Tree )
 
-import Data.Audience
-import Data.AudienceFolder
+
+type Msg
+    = Goto Tree
+    | GotoUp
+
+
+initModel : Model
+initModel =
+    ( buildTree folders audiences, [] )
+
+
+handleDecodedResult : Result String (List a) -> List a
+handleDecodedResult r =
+    case r of
+        Ok results ->
+            results
+
+        Err err ->
+            Debug.log err []
+
+
+audiences : List Audience
+audiences =
+    Json.Decode.decodeString audiencesDecoder Data.Audience.audiencesJSON |> handleDecodedResult
+
+
+folders : List AudienceFolder
+folders =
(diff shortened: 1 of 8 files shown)


caitt:
Interesting challenge.

Due to lack of time and experience with Elm I went for very economic approach:
 - the UI is terrible
 - no tests (decoders and tree building could benefit from some unit tests)
 - did not spent much time on polishing and figuring out what idiomatic Elm code is (maybe Maybe can be handled more elegantly)
 - did not spent much time analyzing performance but it seems straightforward (unless there is something hidden in the libraries) 

From the description I immediately got the basic outline of a solution - use JSON decoders to get list of audiences and folders, then transform it into tree-like structure where one can easily go to parent node (zippers come to mind). Make this structure a model, have two messages (descend to folder X, go to parent). The rest is basic Elm application.

src/Main.elm
```diff
@@ -1,16 +1,96 @@
...
+folders : List AudienceFolder
+folders =
+    Json.Decode.decodeString foldersDecoder Data.AudienceFolder.audienceFoldersJSON |> handleDecodedResult
+
+
+goUpButton : List Tree -> Html Msg
+goUpButton t =
+    case t of
+        x :: xs ->
+            Html.div [ class "goup", onClick (GotoUp) ] [ Html.text "Go UP" ]
```
turboMaCk, line 51: redundant parentheses (not a big deal)

src/Tree.elm
```diff
@@ -0,0 +1,51 @@
...
+        getCurrentFolderData dict =
+            dict |> Dict.get folder.id |> Maybe.withDefault []
+
+        subfolders =
+            getCurrentFolderData subfoldersDict
+
+        leafs =
+            getCurrentFolderData leafDict
+    in
+        Tree folder (List.map (\f -> makeBranch f subfoldersDict leafDict) subfolders) leafs
```
turboMaCk, line 29: this recursion can blow stack with some data. On the other hand it’s quite common issue in situations like this as writing tail recursive variant is incredibely complex (and often has other performance trade offs).

turboMaCk (commented):
First sory it took so long to review this. Unfortunetely I was extremly overloaded with work in past few weeks.

I really like this approach. I think this is the first solution which uses zipper which is at least in my opinione best technique for this kind of thing. I also like that Tree is domain specific one (has folders and audiences). Looks similar to the last chapter of Learn You Haskell if I remember it corectly :+1: Anyway there is one thing this implementation solves differently than I would expect. Usually with this type of tree accumulator is composed from some `BreadCrumb` which contains things on the left and on the right from item so same item is never present twice anywhere in `Zipper` which make certain types of bugs go away. In this case it’s not a big deal though as long as there are no edit actions.

I really like the way logic is separated to modules 👍 

To sum up I think this is best possible approach to something like this despite the fact that I would expect slightly different definition of Zipper. Design of code (modules) is great and scalable (UI decouplet from data manipulations). Very impressive easecially when I take to account that whole implementation is just 251 LOC :+1:

PR #9, opened 2018-07-09
My implementation

Janiczek:
Some notes:

- Applicative style didn't seem to be necessary, as the number of fields in the `Audience` / `AudienceFolder` records fit comfortably into the `Json.Decode.map...map8` range. If I needed more than that, I'd go with [`Json.Decode.Pipeline`](http://package.elm-lang.org/packages/NoRedInk/elm-decode-pipeline/3.0.0/Json-Decode-Pipeline) or [`andMap`](http://package.elm-lang.org/packages/elm-community/json-extra/2.7.0/Json-Decode-Extra#andMap), which seem to amount to the same Applicative abstraction.

- I went from `List Audience` and `List AudienceFolder` to `Dict Int Audience` and `Dict Int AudienceFolder` for the parsed data in my implementation. Could go with RoseTree and/or a zipper for it, which would probably be a bit more performant, but it didn't seem worth it (performance is fine for the current data and it would be a bit more cumbersome to use zipper). I have to `Dict.filter` in the `view` though, which wouldn't be needed with RoseTree.

- The `model.currentFolder` could be something like
```elm
type Location
    = Root
    | Folder Int
```
instead of `Maybe Int` to be a bit more descriptive about the data, but in the end I went with the `Maybe`.

- I factored the parsed data out into top-level constants, don't store them in the model. That would obviously need to happen if I were to fetch the data from the API for real.

### DIFF
--- a/src/Data/Audience.elm
+++ b/src/Data/Audience.elm
@@ -1,12 +1,18 @@
-module Data.Audience exposing (AudienceType(..), Audience, audiencesJSON)
+module Data.Audience exposing (Audience, AudienceType(..), audiencesDecoder, audiencesJSON)
 
 {-| Data.Audiences module
 
 This module implements everything related to audience resource.
 
+
 # Interface
+
 @docs AudienceType, Audience, audienceJSON
+
 -}
+
+import Json.Decode as JD exposing (Decoder)
+
 
 -- Type definition
 
@@ -43,6 +49,7 @@ This is how we usually deal with making non-breaking continuous changes
 from old version of API to new one.
 
 You're free to use any strategy to decode JSON.
+
 -}
 audiencesJSON : String
 audiencesJSON =
@@ -5917,3 +5924,46 @@ audiencesJSON =
         ]
     }
     """
+
+
+
+-- Decoders. Fairly trivial, except for the `audienceTypeDecoder`.
+
+
+audiencesDecoder : Decoder (List Audience)
+audiencesDecoder =
+    JD.field "data" (JD.list audienceDecoder)
+
+
+audienceDecoder : Decoder Audience
+audienceDecoder =
+    JD.map4 Audience
+        (JD.field "id" JD.int)
+        (JD.field "name" JD.string)
+        (JD.field "type" audienceTypeDecoder)
+        (JD.field "folder" (JD.maybe JD.int))
+
+
+{-| I went with looking at the "type" field, as that seemed to have all the
+needed info all in itself. I might be wrong with my reading of the backend
+developers' minds :) But it seems to be the "current" intended way to consume
+the API.
+-}
+audienceTypeDecoder : Decoder AudienceType
+audienceTypeDecoder =
+    JD.string
+        |> JD.andThen
+            (\type_ ->
+                case type_ of
+                    "user" ->
+                        JD.succeed Authored
+
+                    "curated" ->
+                        JD.succeed Curated
+
+                    "shared" ->
+                        JD.succeed Shared
+
+                    _ ->
+                        JD.fail "Unknown AudienceType"
+            )
--- a/src/Main.elm
+++ b/src/Main.elm
@@ -1,16 +1,200 @@
 module Main exposing (main)
 
+import Data.Audience as Audience exposing (Audience)
+import Data.AudienceFolder as AudienceFolder exposing (AudienceFolder)
+import Dict exposing (Dict)
 import Html exposing (Html)
+import Html.Attributes
+import Html.Events
+import Json.Decode as JD exposing (Decoder)
 
 
--- Import Modules
-
-import Data.Audience
-import Data.AudienceFolder
-
-
-{-| Main file of application
+{-| Because of the fixtures, we don't need a full fledged `Html.program`.
 -}
-main : Html msg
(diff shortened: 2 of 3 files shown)


Janiczek:
READY

Janiczek:
Also, I wasn't allowed to change the core types, but I'd think about wrapping the Audience ID and Folder ID into their own types:

```elm
type AudienceId = AudienceId Int
type AudienceFolderId = AudienceFolderId Int
```

This would need some juggling to make it work with a `Dict`, but I think the added safety would be worth it.

src/Main.elm
```diff
@@ -1,16 +1,200 @@
...
+
+
+view : Model -> Html Msg
+view model =
+    let
+        {- This filtering would be unnecessary if we parsed the input data into a RoseTree. -}
+        visibleAudiences : List Audience
+        visibleAudiences =
+            audiences
+                |> Dict.filter (\_ audience -> audience.folder == model.currentFolder)
```
janjelinek, line 107: `always  ((==) model.currentFolder << .folder)`
It's just my obsession  :)

Janiczek: Yeah, @turboMaCk was telling me about GWI writing in pointfree style a lot :)

turboMaCk: especially @janjelinek fall for it :D anyway I think in cases like this one it's just matter of personal taste. For an experienced eye, both have a clear intention and this doesn't enforce cryptic naming like `thing, a, item...` as the code isn't that generic so it makes sense to keep it as it is in my opinion.

src/Data/Audience.elm
```diff
@@ -5917,3 +5924,46 @@ audiencesJSON =
...
     """
+
+
+
+-- Decoders. Fairly trivial, except for the `audienceTypeDecoder`.
+
+
+audiencesDecoder : Decoder (List Audience)
+audiencesDecoder =
+    JD.field "data" (JD.list audienceDecoder)
```
turboMaCk, line 5935: I like that you're not exposing all the stuff :+1:

src/Data/Audience.elm
```diff
@@ -5917,3 +5924,46 @@ audiencesJSON =
...
+        (JD.field "id" JD.int)
+        (JD.field "name" JD.string)
+        (JD.field "type" audienceTypeDecoder)
+        (JD.field "folder" (JD.maybe JD.int))
+
+
+{-| I went with looking at the "type" field, as that seemed to have all the
+needed info all in itself. I might be wrong with my reading of the backend
+developers' minds :) But it seems to be the "current" intended way to consume
+the API.
```
turboMaCk, line 5950: If you're wrong then we have a bug in production:D Anyway I've derived the same thing and think it's the best solution as it prevents states of having `user` and `curated` audience at the same time.

src/Main.elm
```diff
@@ -1,16 +1,200 @@
...
+        { model = init
+        , update = update
+        , view = view
+        }
+
+
+type alias Model =
+    { {- We could use RoseTree and/or a zipper, but I went with the dead-easy
+         way of filtering the dict (or flat list in my first attempt) each time.
+         Might reconsider this if dealing with performance issues!
```
turboMaCk, line 26: :+1:

src/Main.elm
```diff
@@ -1,16 +1,200 @@
...
+         ---
+
+         `Nothing` here means the root folder. We also might create a custom
+         union type to make our intentions more clear:
+
+             type Location = Root | Folder Int
+
+         Some helper functions (`map` + `withDefault` or conversion to `Maybe`
+         come to mind) would then be benefitial.
+      -}
```
turboMaCk, line 37: :+1:

src/Main.elm
```diff
@@ -1,16 +1,200 @@
...
+
+type Msg
+    = OpenFolder (Maybe Int)
+
 
-import Data.Audience
-import Data.AudienceFolder
+{-| We could have this data in model, and in a real app we probably would
+(because this data wouldn't be a constant fixture but something fetched from
+the API). But given the constraints, it's nice to factor this out of the model.
```
turboMaCk, line 48: :+1:

src/Main.elm
```diff
@@ -1,16 +1,200 @@
...
+        |> Maybe.andThen findFolder
+        |> Maybe.map
+            (\folder ->
+                let
+                    {- Experimented with a few ways to get both the current
+                       folder and the parent folder name, and this `let` inside
+                       the inner function was what I ended up with.
+
+                       It can be juggled to look nicer but perform duplicate
+                       computations. ¯\_(ツ)_/¯ Again, depends on team code style.
```
turboMaCk, line 138: seems good as it is to me :+1:

src/Main.elm
janjelinek: True true, I'm becoming composition and pointfree monster. To be honest it didn't help to readability too much in many times, but you know, my brain sees that automatically, lkie that aiblity to raed wrods with reodrered letters :D

turboMaCk (commented):
Everything looks good to me. I especially appreciate all the comments and description of PR. It's nice to see you've considered various different way how something like this can be done and have captured possible future evolvements for maintainers while choosing a simpler approach that satisfies requirements :+1:

PR #10, opened 2018-09-22
Basic implementation of step 1 - READY

jhrcek:
Hello @turboMaCk,
Here's my first attempt at implementation. I consider this READY for review.

How to run this:
1. `elm make src/Main.elm`
2. Open the generated index.html in the browser

I tried to keep the implementation simple (no external HTML, CSS and minimal dependencies) while still implementing all the requirements.

> How to make JSON parsing scalable? Do you know about applicatives?

Yes I know applicatives :-)
Among other things they're useful when there's more than 8 fields to decode in an object.
But in this case it wasn't necessary, because we needed to decode 3/4 fields, which is
easily handled using just `Json.Decode.mapN` from the `core` package.

> Is there any clever transformation of data that can be used as input of browser itself?

I transformed the flat lists of `AudienceFolder`s and `Audience`s by building a navigation tree out of it and then wrapping it into a zipper. Details below..

> What data-structures can be helpful if any?
> Do you prefer working with RoseTree or List? Why? (you can leave comment in your code)

I think the tree zipper data structure is ideal for this use case.
Although the amount of data provided in the fixtures could be easily handled using plain `List` or rose trees, I wanted to try using `Zipper`.
This made the initial construction of the navigation tree a bit harder, but should have the advantage of good performance even with large data. It also makes the implementation of the navigation simple and elegant (using only single call to `up`, `open`, `children` and `isRoot` respectively from the `Lazy.Tree.Zipper`).
I'm aware of one problem with this implementation: the library (lazy-tree-with-zipper)[https://package.elm-lang.org/packages/turboMaCk/lazy-tree-with-zipper/] I for this is using functions to achieve laziness of the data structure. This could cause [potential problems when debugging](https://groups.google.com/forum/#!topic/elm-discuss/bOAHwSnklLc) (as the debugger can't serialize the zipper state) and goes against the general recommendation "Keep the functions out of your model" .

> Which parts can be decoupled for possible re-use?
> Do you like to put those in separate module? Why yes or no?

I chose to move the navigation tree implementation into separate module and made `TreeItem` into opaque type to hide the implementation from other modules. Apart from reuse, the most common reasons for hiding implementation details is prevent clients from depending on them (thus enabling possible change of implementation in the future) and making the module API simpler (not exposing unnecessary details).

### DIFF
--- a/src/Data/Audience.elm
+++ b/src/Data/Audience.elm
@@ -1,4 +1,11 @@
-module Data.Audience exposing (AudienceType(..), Audience, audiencesJSON)
+module Data.Audience
+    exposing
+        ( Audience
+        , AudienceType(..)
+        , audiences
+        , audiencesDecoder
+        , audiencesJSON
+        )
 
 {-| Data.Audiences module
 
@@ -7,6 +14,9 @@ This module implements everything related to audience resource.
 # Interface
 @docs AudienceType, Audience, audienceJSON
 -}
+
+import Json.Decode as Decode exposing (Decoder)
+
 
 -- Type definition
 
@@ -27,6 +37,52 @@ type alias Audience =
     , type_ : AudienceType
     , folder : Maybe Int
     }
+
+
+audiencesDecoder : Decoder (List Audience)
+audiencesDecoder =
+    Decode.field "data" <| Decode.list audienceDecoder
+
+
+audienceDecoder : Decoder Audience
+audienceDecoder =
+    Decode.map4 Audience
+        (Decode.field "id" Decode.int)
+        (Decode.field "name" Decode.string)
+        (Decode.field "type" audienceTypeDecoder)
+        (Decode.field "folder" <| Decode.nullable Decode.int)
+
+
+audienceTypeDecoder : Decoder AudienceType
+audienceTypeDecoder =
+    Decode.string
+        |> Decode.andThen
+            (\str ->
+                case str of
+                    "user" ->
+                        -- None of the objects in the provided JSON has "type": "authored".
+                        -- Assuming Authored corresponds to "type": "user"
+                        Decode.succeed Authored
+
+                    "shared" ->
+                        Decode.succeed Shared
+
+                    "curated" ->
+                        Decode.succeed Curated
+
+                    _ ->
+                        Decode.fail <| "Invalid AudienceType : " ++ str
+            )
+
+
+{-| Taking advantage of hardcoded data.
+If we were to retrieve the data from the backend, we'd have to manage possible errors,
+store errors in the model to possibly display them in the UI etc., perhaps using remote-data package
+-}
+audiences : List Audience
+audiences =
+    Decode.decodeString audiencesDecoder audiencesJSON
+        |> Result.withDefault []
 
 
 
--- a/src/Main.elm
+++ b/src/Main.elm
@@ -1,16 +1,102 @@
 module Main exposing (main)
 
-import Html exposing (Html)
-
-
 -- Import Modules
 
-import Data.Audience
-import Data.AudienceFolder
+import Data.Audience exposing (audiences)
+import Data.AudienceFolder exposing (audienceFolders)
+import Html exposing (Attribute, Html, text)
+import Html.Attributes exposing (style)
+import Html.Events exposing (onClick)
+import Lazy.Tree.Zipper as Zipper exposing (Zipper)
+import NavTree exposing (TreeItem, isFolderWithId)
 
 
 {-| Main file of application
 -}
-main : Html msg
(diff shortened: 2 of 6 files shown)


src/Data/Audience.elm
```diff
@@ -29,6 +39,52 @@ type alias Audience =
...
+
+audienceTypeDecoder : Decoder AudienceType
+audienceTypeDecoder =
+    Decode.string
+        |> Decode.andThen
+            (\str ->
+                case str of
+                    "user" ->
+                        -- None of the objects in the provided JSON has "type": "authored".
+                        -- Assuming Authored corresponds to "type": "user"
```
turboMaCk, line 64: correct

src/Data/Audience.elm
```diff
@@ -29,6 +39,52 @@ type alias Audience =
...
+
+                    _ ->
+                        Decode.fail <| "Invalid AudienceType : " ++ str
+            )
+
+
+{-| Taking advantage of hardcoded data.
+If we were to retrieve the data from the backend, we'd have to manage possible errors,
+store errors in the model to possibly display them in the UI etc., perhaps using remote-data package
+-}
```
turboMaCk, line 81: :+1:

src/Main.elm
```diff
@@ -1,16 +1,102 @@
...
+    Zipper TreeItem
+
+
+update : Msg -> Model -> Model
+update msg zipper =
+    case msg of
+        GoUp ->
+            Zipper.up zipper
+                -- Nothing means we're already at the root, OK to stay there
+                |> Maybe.withDefault zipper
```
turboMaCk, line 46: This is fine in this case anyway Zipper has helper function `attempt` so less verbose equivalent would be `Zipper.attempt Zipper.up zipper`

src/Main.elm
```diff
@@ -1,16 +1,102 @@
...
+        OpenFolder folderId ->
+            Zipper.open (isFolderWithId folderId) zipper
+                |> Maybe.withDefault zipper
+
+
+view : Model -> Html Msg
+view model =
+    let
+        openedFolderItems =
+            NavTree.sortItems <| Zipper.children model
```
turboMaCk, line 57: alternatively, you can use `Zipper.openAll` which returns a list of the zipper (for children). This has an advantage that then in itemView you would have access to parents etc.

src/Main.elm
```diff
@@ -1,16 +1,102 @@
...
+                []
+
+        goUpButtonIfNotAtRoot =
+            if Zipper.isRoot model then
+                []
+            else
+                [ navigationItem "⇦ Go Up" "gray" [ onClick GoUp ] ]
+    in
+    Html.ul [ style [ ( "font-family", "sans-serif" ) ] ]
+        (goUpButtonIfNotAtRoot
```
turboMaCk, line 72: I would probably prefer `<|` here for esthetic rasons

src/NavTree.elm
```diff
@@ -0,0 +1,113 @@
...
+root : TreeItem
+root =
+    Folder
+        -- ASSUMPTION: all nodes from the backend have positive IDs.
+        -- If this wasn't the case (check with backend team)
+        -- we could find a minimal ID in all the nodes and set the root ID to "min - 1"
+        { id = -1
+        , name = "ROOT"
+        , parent = Nothing
+        }
```
turboMaCk, line 90: alternatively, you can add special `Root` constructor to `TreeItem`

turboMaCk (commented):
Nice work :clap: I tried to comment few things anyway just to give at least some feedback :D

Anyway, this is a very nice implementation IMO. I want to highlight the module organization especially and also the usage of expose and quantification the way I personally like as well. And of course the usage of the zipper which is probably the most advanced technique how to implement something like this.

src/Main.elm
jhrcek: Good tip, didn't notice that exists. I like reading the docs before I start using new package, but in this case I failed to finish reading everything :)

src/NavTree.elm
jhrcek: That's would be possible, but it would open doors to invalid states with Root items at non-root positions in the tree which might be problematic..

jhrcek:
Thanks for the review and useful feedback :)

src/Main.elm
turboMaCk: yeah the surface area of that package is quite big :/

src/NavTree.elm
turboMaCk: true but this problem already exists with this implementation as the folder can exist under the audience.

jhrcek:
Closing this as I don't like keeping PRs opened that will never be merged :stuck_out_tongue:

PR #13, opened 2019-06-01
Matus Ledenyi - Challenge

mathiajusth:
Hi!
I finally arived from Portugal and finished the challenge.

Notes: 
- I used `rtfeldman/elm-css` for rendering, I am not sure whether this goes against using `elm/html` as I am not acquainted with the purpose of this restriction. To be safe, I contained it in one module and I did not use any third party rendering functions. This can be of course reworked if you would like to recommend me your preferred way of using css.
- as icons I just used some online resource so internet connection is needed
- I took an approach of designing custom data type for the file system. The development time could surely be shortened by using some tree zipper library

Thank you for reviewing this and have a nice weekend


mathiajusth:
READY

src/Json/Decode/Applicative.elm
```diff
@@ -0,0 +1,19 @@
+module Json.Decode.Applicative exposing (apply, with)
+
+{-| Json.Decode.Applicative module
+
+This module implements typeclass applicative (apply : m (a -> b) -> a -> b)
```
jhrcek: I guess you mean `m (a -> b) -> m a -> m b` :-)

src/Data/Audience.elm
```diff
@@ -36,6 +41,48 @@ type alias Audience =
...
+            D.succeed Shared
+
+        "user" ->
+            D.succeed Authored
+
+        type_ ->
+            D.fail <|
+                "Invalid audience type: "
+                    ++ type_
+                    ++ "! Only \"curated\", \"shared\" and \"user\" are valid."
```
jhrcek, line 63: Nice. I like it when error messages are informative :-)

.gitignore
```diff
@@ -1 +1,6 @@
 elm-stuff/
+
+/* vim backup files */
+
+*.swp
+*.swo
```
turboMaCk: I would recommend something like https://github.com/turboMaCk/Dotfiles/blob/master/vimrc#L219-L224 or using `~/.gitignore_global` for this. Just a tip to save you from doing this over and over again.

src/Json/Decode/Applicative.elm
```diff
@@ -0,0 +1,19 @@
...
+for Decoder functor based on the monadic return (succeed) and bind (andThen)
+
+-}
+
+import Json.Decode exposing (Decoder, andThen, succeed)
+
+
+apply : Decoder (a -> b) -> Decoder a -> Decoder b
+apply df da =
+    andThen (\a -> andThen (\f -> succeed (f a)) df) da
```
turboMaCk: this is fine but my favorite way of defining applicative map is `map2 (|>)` which is ridiculously simple.

src/Json/Decode/Applicative.elm
```diff
@@ -0,0 +1,19 @@
...
+
+
+with da df =
+    apply df da
```
Janiczek: What are your thoughts on `elm-community/json-extra` and https://package.elm-lang.org/packages/elm-community/json-extra/latest/Json-Decode-Extra#andMap ?

turboMaCk: I can see the order of args in apply is probably more common across languages but in elm, I would probably define it like this `flip apply` version right away.

elm.json
```diff
@@ -8,17 +8,23 @@
         "direct": {
             "elm/browser": "1.0.1",
             "elm/core": "1.0.2",
-            "elm/html": "1.0.0"
+            "elm/html": "1.0.0",
+            "elm/json": "1.1.3",
+            "elm-community/list-extra": "8.2.0",
+            "elm-explorations/test": "1.2.1",
```
turboMaCk: shouldn't be this in `test-dependencies` 🤔 ?

src/Data/FileSystem.elm
```diff
@@ -0,0 +1,186 @@
...
+mapFiles f tree =
+    case tree of
+        File x ->
+            File (f x)
+
+        Folder name content ->
+            Folder name (List.map (mapFiles f) content)
+
+
+mapFolders : (List (FileSystem a) -> List (FileSystem a)) -> FileSystem a -> FileSystem a
```
turboMaCk: this is functor probably but in a different category than one of set and functions. I would probably use `update` name as it's `a -> a`

src/Data/FileSystem.elm
```diff
@@ -0,0 +1,186 @@
...
+                (content
+                    ++ List.map
+                        (\subFolder ->
+                            mkFileSystemHelper
+                                restFolders
+                                (\af -> af.parent == Just subFolder.id)
+                                restFiles
+                                (\a -> a.folder == Just subFolder.id)
+                                (Folder subFolder.name [])
+                        )
```
turboMaCk, line 143: Q: What do you think about the complexity of this? (no need for a mathematical answer, it can be just practical / gut feeling)

src/Data/List/Utility.elm
```diff
@@ -0,0 +1,16 @@
+module Data.List.Utility exposing (splitFilter)
+
+
+splitFilter : (a -> Bool) -> List a -> ( List a, List a )
```
jhrcek: This function already exists in `elm/core`:  https://package.elm-lang.org/packages/elm/core/latest/List#partition, please reuse that.

turboMaCk (commented):
Sorry I was on holiday last week as well so that's why it took so long.

This looks great 🙌 Probably my primary concern would be naming. In particular, I would probably avoid using "filesystem" analogy even though it's indeed similar to that ([ZipperFS](https://github.com/jkarni/ZipperFS) yeah). But that's really just cosmetic thing. I've tried to comment on a few things anyway to mostly let you know that I indeed read the code 😂 .

src/Data/FileSystem.elm
```diff
@@ -0,0 +1,186 @@
...
+    mapFolders List.reverse
+
+
+sortWith : (FileSystem a -> FileSystem a -> Order) -> FileSystem a -> FileSystem a
+sortWith sorter =
+    mapFolders (List.sortWith sorter)
+
+
+sortFoldersAlphabetically : FileSystem a -> FileSystem a
+sortFoldersAlphabetically =
```
jhrcek: This function is probably not needed, because `sortAudienceFilesAndFoldersAlphabetically` does all we need.

src/Data/FileSystem.elm
```diff
@@ -0,0 +1,186 @@
...
+                        EQ
+
+
+{-| Interface with BE Audience & AudienceFolder
+
+mkFileSystem transforms BE data into FileSystem of Audiences
+
+-}
+mkFileSystem : List AudienceFolder -> List Audience -> FileSystem Audience
+mkFileSystem folders files =
```
kraklin: What is the `mk` prefix stands for? I'd either improve or explaine the choice of that prefix in the comment.

turboMaCk: usually, it means `make`

jhrcek (changes requested):
Nice work! I see you have some experience with FP and some Css skills too :smiley: 
I just have one problem with the solution: It doesn't show any Folders containing subfolders.
Notice, that in the test data there is folder "New Group" with id 3110 which contains another folder "New Group 2" with id 3111. But It's not displayed in your solution. Can you try to fix your solution so this folder is also shown and can be opened?

src/Data/FileSystem.elm
turboMaCk: I'm not even sure from where I know this convention. Is it from Haskell or Rust or Nix or something else?

kraklin: I'd go with `makeFileSystem` then.

jhrcek: mk stands for make and is usually used for "dumb" constructor-like functions. It's relatively popular in Haskell (it's used all over the place in ghc compiler)

src/Json/Decode/Applicative.elm
turboMaCk: I think we should consider that the primary goal of this implementation is to show off an understanding of things and thus I would even appreciate "reinventing of the wheel".

src/Data/FileSystem.elm
```diff
@@ -0,0 +1,186 @@
...
+                        subFolders
+                    ++ List.map File subFiles
+                )
+
+
+sortAudienceFilesAndFoldersAlphabetically : FileSystem Audience -> FileSystem Audience
+sortAudienceFilesAndFoldersAlphabetically =
+    sortWith <|
+        \fs1 ->
+            \fs2 ->
```
Janiczek: I've seen this pattern a few times here - just a note you can use
```elm
sortWith <|
  \fs1 fs2 ->
    case ...
```
or even, for readability:
```elm
let
    sortingFn fs1 fs2 =
        case ...
in
    sortWith sortingFn
```
(although I like the first example better)

src/Data/FileSystem.elm
turboMaCk: in elm usual convention is to use `init` over `make` I would say. But that's just another cosmetic thing.

elm.json
Janiczek: It's an `elm install` bug, mostly. Some packages regularly depend on `elm-explorations/test` and then `elm make` flips over and refuses to do anything, and you have to remove `test` from the regular deps, and only have it in the test deps. Happens to our codebase almost every time somebody touches elm.json.

src/Main.elm
```diff
@@ -1,10 +1,110 @@
...
+    Result String AudienceBrowser.Model
+
+
+
+-- INIT
+
+
+init : flags -> ( Model, Cmd msg )
+init =
+    \_ ->
```
Janiczek: cosmetic stuff:
```elm
init _ = ...
```

src/Json/Decode/Applicative.elm
mathiajusth: I certainly could have used that but as @turboMaCk wrote I treated this as an exercise so I wanted to implement the applicative myself (I have actually never implemented Aplicative using Monad so it was a nice exercise for me).

.gitignore
mathiajusth: thanks!

src/Json/Decode/Applicative.elm
turboMaCk: we actually do have one implementation of the applicative using monad in our production code even. It's used to implement permutations of values of `List a, List b` to `List (a,b)` :)

turboMaCk: for curious souls (it's not 100% just an applicative).

```elm
allCacheKeysFor : List DatapointCode -> List AudienceId -> List SegmentId -> List CacheKey
allCacheKeysFor datapointCodes audienceIds segmentIds =
    let
        withMaybe : List a -> (Maybe a -> b) -> List b
        withMaybe xs f =
            if List.isEmpty xs then
                List.singleton <| f Nothing

            else
                List.map (f << Just) xs
    in
    List.map CacheKey datapointCodes
        |> List.andThen (withMaybe audienceIds)
        |> List.andThen (withMaybe segmentIds)
```

so I think seeing you understand similar patterns is valuable for us.

src/Json/Decode/Applicative.elm
mathiajusth: That is certainly more insightful.
I took this as an exercise of defining Applicative in terms of Monad and havent even thought of this neat way.

mathiajusth:
@jhrcek 
> I just have one problem with the solution: It doesn't show any Folders containing subfolders.

I was [not showing](https://github.com/mathiajusth/challenge-elm/blob/8c81fae5b39eb11102c6a1f1b8eb8b95435aab28/src/View/AudienceBrowser.elm#L351-L356)
 folders which did not contain any files. I'll remove this filter.

src/Data/FileSystem.elm
mathiajusth: I haven't really done any complexity theory yet, but I'd say:
I have to filter through not more than all files and folders for each folder + 1 (the root folder) so
```
let f := length files 
    F := length folders
in (f+F)*(F+1) ~> F^2 + f*F + F + f
```
So in terms of the `n := length (files ++ folders)` considering that `n = f + F` hence `f < n` & `F < n` it is better than `n^2 + n^2 + 2n` ~> `2*n^2 + 2*n`

I guess the answer is at worst quadratic and for shallow folder structures much better (close to linear? - for a fixed folder structure it is linear `(F+1)*f + (F^2 + F)`).

src/Data/FileSystem.elm
mathiajusth: Thx, I like the first one the most, as well.

mathiajusth:
Thank you for your kind reviews. 
Your suggestions have been implemented.

jhrcek (approved):
Great work. Looks good to me.

turboMaCk:
:grimacing: sorry for that. I know the provided data are not ideal - my bad actually.

PR #14, opened 2019-06-24
First & second step READY

annaghi:
Hello @turboMaCk ,

this is my implementation, it is READY for review.

I explain my thoughts in the `NOTES.md`.

You can run the app with an http server, the config can be found in the `package.json`.
You can install the dev-dependencies with:
```bash
npm install
```
and run the app with:
```bash
npm run watch
```
Regards,
Anna

### DIFF
--- a/FIRST_STEP.md
+++ b/FIRST_STEP.md
@@ -8,36 +8,44 @@ Your job is to implement parsing of data from `JSON` format and fully usable int
 This is required for successful implementation:
 
 - [ ] You must use data and type (aliases) provided.
-- [ ] You must use provided `JSON` fixtures in your app (*these are returned by API in production app*)
+- [ ] You must use provided `JSON` fixtures in your app (_these are returned by API in production app_)
 - [ ] You must not change any of pre-defined types and fixtures.
-    - However you can use existing files for your implementation.
+  - However you can use existing files for your implementation.
 - [ ] You must successfully resolve all relations in data set.
 - [ ] By default component should display root level (items without any parent folder)
 - [ ] Only items from current level are shown (in UI) at any given time.
 - [ ] Folders are displayed before audiences in any level.
 - [ ] Click on any folder opens its content.
-- [ ] `Go up` button is not visible in root level.
-- [ ] `Go up` button is visible in any sub-level.
-    - You can use name of parent folder if you want.
-- [ ] `Go up` button opens parent level on click.
+- [x] `Go up` button is not visible in root level.
+- [x] `Go up` button is visible in any sub-level.
+  - You can use name of parent folder if you want.
+- [x] `Go up` button opens parent level on click.
 - [ ] Final solution should be compilable to html app or should contain HTML and source compilable to JavaScript included in HTML.
+
+#### Proposal
+
+My proposal is to replace the `Go up` button related requirements with the following ones:
+
+- [ ] An `Audiences` breadcrumb is visible on root level.
+- [ ] Display the breadcrumbs of the current level.
+- [ ] Click on any ascendant breadcrumb opens the corresponding level (navigates up/upwards in the tree of folders).
 
 ## Hints
 
 These are the things you might consider thinking about before you start with your implementation:
 
 - How to make `JSON` parsing scalable?
-    - Do you know about [applicatives](https://toast.al/posts/2016-08-12-elm-applicatives-and-json-decoders.html)?
+  - Do you know about [applicatives](https://toast.al/posts/2016-08-12-elm-applicatives-and-json-decoders.html)?
 - Is there any clever transformation of data that can be used as input of browser itself?
-    - What data-structures can be helpful if any?
-    - Do you prefer working with RoseTree or List? Why? (you can leave comment in your code)
+  - What data-structures can be helpful if any?
+  - Do you prefer working with RoseTree or List? Why? (you can leave comment in your code)
 - Which parts can be decoupled for possible re-use?
-    - Do you like to put those in separate module? Why yes or no?
+  - Do you like to put those in separate module? Why yes or no?
 
 ## Final Words
 
 Please don't be afraid to ask (using email or by opening issue in this repository) if anything is not clear,
 can be improved in this documents or if you're just simply stuck with anything.
 We are happy to provide any help to degree we are able to. Also have in mind that there are no time restrictions for completing this challenge.
-Take as much time as you wish for completing this. *In the end the most important bit is to challenge your problem solving thinking*.
+Take as much time as you wish for completing this. _In the end the most important bit is to challenge your problem solving thinking_.
 And of course **good luck with this**! :rocket:
--- a/NOTES.md
+++ b/NOTES.md
@@ -1,0 +1,66 @@
+### Problem domain
+
+The challenge reminds me of the [Faceted Search](https://en.wikipedia.org/wiki/Faceted_search). Our **data** is the list of audiences and we have three **filters** on it:
+
+1. a hierarchical filter with the tree of folders,
+2. a categorical filter with audience types, and
+3. a categorical (or ordinal) filter with audience and folder names.
+
+These filters have different representations on the screen:
+
+1. browser (first step)
+2. three buttons (second step)
+3. search box (on the screenshot)
+
+According to the faceted search, **applying filters on data** may mean:
+filteredData = ⋂ filter<sub>i</sub> (data)
+
+### Proposal
+
+I think this is a scalable, robust technique. And so I would like to propose some changes in the requirements. I have added the proposed changes to the `FIRST_STEP.md` and the `SECOND_STEP.md`, and the Elm code is a prototype implementation of those changes.
+
+I know that this challenge is about meeting the requirements, but perhaps my contribution may be beneficial too.
+
+#### First step
+
+In the implementation I have used the lazy rose tree package, and the hierarchical filter was basicly done.
+The data is baked into the hierarchy itself, so applying **another hierarchical filter** would not be easy right now. I think this is the only limitation of my implementation in contrast of the faceted search technique.
+
+#### Second step
+
+In order to implement a (non-hierarchical) filter I have created a mutual recursive function-pair for the lazy rose tree package. With that we can decide whether an audience should be displayed or not in the `levelView` function.
+
+Adding more filters will be straightforward, we need to implement the intersection in the `checker` helper functions.
+
+- Two new functions in the `Lazy.Tree` called `isInTree` and `isInForest`, and
+- a wrapper function in the `Lazy.Tree.Zipper` called `isInZipper`.
+
+References for this step:
+
+- [Wikipedia - Mutual recursion](https://en.wikipedia.org/wiki/Mutual_recursion)
+- [Simply Scheme - Trees](https://people.eecs.berkeley.edu/~bh/pdf/ssch18.pdf) pp. 310 - 313
+
+#### Design
+
+I would like to explain my design choices using user stories (if I were a user).
+
+As a user I want to see breadcrumbs so that I do not need to remember what path led to these audiences.
+This is why I have added breadcrumbs at the top of the list.
+
+As a user I want to open a folder with a single click so that I can reach my data quickly.
+This is why the breadcrumbs are clickable. Also in order to be consistent, the user always clicks on a folder, no matter if they traverse up or down in the hierarchy.
+
+As a user I want to know how I had filtered my audiences out so that I can easily pick up my thoughts after a coffee-break.
+This is why the filters' current state is upfront. The users can capture only the red elements on the sidebar and recall all their settings.
+
+### Potential benefits
+
+- The browser can be extracted into its own reusable module.
+
+- All the non-hierarchical filter types can be extracted into their own reusable modules.
+
+- Adding new filters can be implemented with the same pattern.
+
+- Whenever a user sees a browser+filters widget on the site, they know that it works consistently.
+
+- The users can organize their data into folders as they want, because the folders are neutral.
--- a/SECOND_STEP.md
+++ b/SECOND_STEP.md
@@ -22,14 +22,20 @@ Categories are:
 ### These are the requirements for this step:
 
 - [ ] You must not change any of pre-defined types and fixtures.
-    - However you can use existing files for your implementation.
+  - However you can use existing files for your implementation.
 - [ ] Add 3 buttons - one for every category of audiences.
 - [ ] Clicking on each button activates particular category of audience.
 - [ ] By default `Authored` audiences are filtered.
 - [ ] There must by at least one category selected at the time.
-- [ ] Every-time user changes category she sees its root level.
-- [ ] In case of shared audiences ignore tree structure and display audiences in single level without any folder.
+- [x] Every-time user changes category she sees its root level.
+- [x] In case of shared audiences ignore tree structure and display audiences in single level without any folder.
 - [ ] Final solution should be compilable to html app or should contain HTML and source compilable to JavaScript included in HTML.
+
+#### Proposal
+
+My proposal is to replace the two checked requirements with the following one:
+
+- [ ] Every-time user changes category she sees the audiences of the selected categories.
 
 ## Hints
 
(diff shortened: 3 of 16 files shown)


turboMaCk:
Thanks a lot @annaghi I and team are going to review this ASAP!

src/Data/Audience.elm
```diff
@@ -35,6 +42,129 @@ type alias Audience =
...
+
+
+audiencesDecoder : Json.Decode.Decoder (List Audience)
+audiencesDecoder =
+    Json.Decode.field "data" (Json.Decode.list audienceDecoder)
+
+
+audienceDecoder : Json.Decode.Decoder Audience
+audienceDecoder =
+    Json.Decode.map4 Audience
```
kraklin, line 76: Using `Decode.mapN` is OK, but in my eyes the `mapN` is limited on large JSON objects (I think `map8` is the largest one). I'd prefer `Json.Decode.Pipeline` instead.

src/Data/Audience.elm
```diff
@@ -35,6 +42,129 @@ type alias Audience =
...
+        "user" ->
+            Ok Authored
+
+        "shared" ->
+            Ok Shared
+
+        "curated" ->
+            Ok Curated
+
+        unknown ->
```
kraklin, line 95: Wow, love the usage of `unknown` instead of `_` and having the problem in the Error. Nice thinking. 👍

src/Main.elm
```diff
@@ -1,10 +1,405 @@
...
+            [ Html.Attributes.classList [ ( "selected", isSelected ) ]
+            , Html.Attributes.style "cursor" "pointer"
+            , Html.Events.onClick (AudenceTypeToggled audienceType (not isSelected))
+            ]
+            [ Html.i [] [ audienceTypeInfo.icon ]
+            , Html.text audienceTypeInfo.text
+            ]
+
+
+errorView : List String -> List (Html.Html Msg)
```
kraklin, line 346: 😍 Love the error view with encouraging message to open issue.

kraklin (commented):
🚀 Thank you for the great PR. I hope you have enjoyed the challenge. There is not much to question from my point of view after first quick sweep of your code (except of suggestion of usage `Json.Decode.Pipeline` instead of `map`. 

I like the structure, explanations of you thinking and those docs (and doctests) of Lazy library are excellent.
Good job

src/Data/Audience.elm
Janiczek: @kraklin I believe I did use `map4` in my implementation of the challenge too - because for this small example it's enough, and the type won't grow (unlike in real apps :upside_down_face: ).

src/Data/Audience.elm
```diff
@@ -35,6 +42,129 @@ type alias Audience =
...
+        |> List.indexedMap Tuple.pair
+        |> List.foldl
+            (\( index, actualVariant ) acc ->
+                if actualVariant == variant then
+                    index
+
+                else
+                    acc
+            )
+            -1
```
Janiczek, line 165: So, this is kinda like [`List.Extra.elemIndex`](https://package.elm-lang.org/packages/elm-community/list-extra/latest/List-Extra#elemIndex), right? (well, that and `Maybe.withDefault -1`)

I think I'd rather use an existing function -- but I understand that bringing big libraries into an example project would seem icky. (We do use them in the real codebase though so maybe we should encourage people doing the challenge to feel free to do so too :thinking: )

src/Lazy/Tree.elm
```diff
@@ -0,0 +1,618 @@
...
+    --> True
+
+    Tree 1 (LL.fromList [ insert (singleton 5) <| singleton 2, insert (singleton 6) <| singleton 3, singleton 4 ])
+        |> isInTree ((==) 4)
+    --> True
+
+-}
+isInTree : (a -> Bool) -> Tree a -> Bool
+isInTree predicate (Tree value forest) =
+    predicate value || isInForest predicate forest
```
Janiczek, line 551: Highlighting the additions to the vendored libraries 1/3

src/Lazy/Tree.elm
```diff
@@ -0,0 +1,618 @@
...
+    --> True
+
+    Tree 1 (LL.fromList [ insert (singleton 5) <| singleton 2, insert (singleton 6) <| singleton 3, singleton 4 ])
+        |> descendants
+        |> isInForest ((==) 4)
+    --> True
+
+-}
+isInForest : (a -> Bool) -> Forest a -> Bool
+isInForest predicate forest =
```
Janiczek, line 595: Highlighting the additions to the vendored libraries 2/3

src/Lazy/Tree/Zipper.elm
```diff
@@ -0,0 +1,737 @@
...
+
+    T.Tree 1 (LL.fromList [ T.insert (T.singleton 5) <| T.singleton 2, T.insert (T.singleton 6) <| T.singleton 3, T.singleton 4 ])
+        |> fromTree
+        |> isInZipper ((<) 2)
+    --> True
+
+-}
+isInZipper : (a -> Bool) -> Zipper a -> Bool
+isInZipper predicate (Zipper t bs) =
+    Tree.isInTree predicate t
```
Janiczek, line 679: Highlighting the additions to the vendored libraries 3/3

src/Main.elm
```diff
@@ -1,10 +1,405 @@
...
+
+
+type Node
+    = Audience Data.Audience.Audience
+    | AudienceFolder Data.AudienceFolder.AudienceFolder
+
+
+type alias Model =
+    -- An item of the `zipper` stores whether it is open, whether it is selected, and the audience/folder properties.
+    { zipper : Lazy.Tree.Zipper.Zipper ( Bool, Bool, Node )
```
Janiczek, line 41: I'd prefer either some custom types to give the bools a name (mumble mumble _boolean identity crisis_ mumble mumble), or changing this 3-tuple into a record. I know "what is what" is described in the comments but records seem better to me in that regard.

src/Data/Audience.elm
jhrcek, line 165: Returning `-1` for values seems kind of arbitrary. What if someone later uses this function without realizing it can return negative values? Can you think of a simpler way to associate integer index with each possible value of `AudienceType`?

Janiczek:
Very interesting wrt. the faceted search. Also, other than some nitpicks (and some typos which I didn't flag here as I think there's no point) I don't have much to say - the code looks great! Kudos for making sense of @turboMaCk's lazy-tree-with-zipper library and even extending it :smile:

src/Lazy.elm
```diff
@@ -0,0 +1,202 @@
+module Lazy exposing
```
jhrcek, line 1: This code seems to be copy-pasted from [lazy-tree-with-zipper](https://package.elm-lang.org/packages/turboMaCk/lazy-tree-with-zipper/3.1.1/). What was your reasoning behind copy-pasting the code (as opposed to depending on that package in the elm.json)?

jhrcek (commented):
Nice solution! Visually your solution is among the best ones I've seen. I also like the breadcrumb path implementation for easier navigation :+1: Just a couple questions/comments inline..

src/Lazy.elm
Janiczek: @jhrcek Additional functions needed to be implemented: https://github.com/GlobalWebIndex/challenge-elm/pull/14/files#diff-c7375ff11bc41bd2c1fbde6f98f5904aR35

src/Data/Audience.elm
```diff
@@ -35,6 +42,129 @@ type alias Audience =
...
+        Shared ->
+            AudienceTypeInfo index "Shared" Icons.share2
+
+        Curated ->
+            AudienceTypeInfo index "Default" Icons.list
+
+
+getIndexByVariant : v -> List v -> Int
+getIndexByVariant variant list =
+    list
```
janjelinek, line 155: It's nice how simply extensible it is, but for this case seems overcomplicated to me.

You can simply remove this whole function and above change `case` to:
```
case audienceType of
    Authored ->
        AudienceTypeInfo 0 "My audiences" Icons.user

    Shared ->
        AudienceTypeInfo 1 "Shared" Icons.share2

    Curated ->
        AudienceTypeInfo 2 "Default" Icons.list
```
and it should be enough because I didn't find any relying on order (index from defined order in the list) in your code.

janjelinek (commented):
I really like the approach to errors and error messages. Rest of important was already written by others and I didn't find anything else. 👍

FIRST_STEP.md
```diff
@@ -8,36 +8,44 @@ Your job is to implement parsing of data from `JSON` format and fully usable int
...
 - [ ] Click on any folder opens its content.
-- [ ] `Go up` button is not visible in root level.
-- [ ] `Go up` button is visible in any sub-level.
-    - You can use name of parent folder if you want.
-- [ ] `Go up` button opens parent level on click.
+- [x] `Go up` button is not visible in root level.
+- [x] `Go up` button is visible in any sub-level.
+  - You can use name of parent folder if you want.
+- [x] `Go up` button opens parent level on click.
 - [ ] Final solution should be compilable to html app or should contain HTML and source compilable to JavaScript included in HTML.
```
turboMaCk, line 23: this one is also done btw. All we meant by that is that we're able to run it locally without writing some HTML and JS wrapping code around it. I was able to run it locally without any issues.

FIRST_STEP.md
```diff
@@ -8,36 +8,44 @@ Your job is to implement parsing of data from `JSON` format and fully usable int
...
 
+#### Proposal
+
+My proposal is to replace the `Go up` button related requirements with the following ones:
+
+- [ ] An `Audiences` breadcrumb is visible on root level.
+- [ ] Display the breadcrumbs of the current level.
+- [ ] Click on any ascendant breadcrumb opens the corresponding level (navigates up/upwards in the tree of folders).
+
```
turboMaCk, line 32: this sounds good. We're actually using something like this in many places just not this one. The reason is that in actual production version as we have at the moment we allow just a single level of audiences folders. For the purpose of this challenge I made it multi-level (like other more complicated things we have) to make it more interesting but I haven't changed the "breadcrumbs" behavior. But it's nice to see you're thinking about UX :)

NOTES.md
```diff
@@ -0,0 +1,66 @@
...
+The data is baked into the hierarchy itself, so applying **another hierarchical filter** would not be easy right now. I think this is the only limitation of my implementation in contrast of the faceted search technique.
+
+#### Second step
+
+In order to implement a (non-hierarchical) filter I have created a mutual recursive function-pair for the lazy rose tree package. With that we can decide whether an audience should be displayed or not in the `levelView` function.
+
+Adding more filters will be straightforward, we need to implement the intersection in the `checker` helper functions.
+
+- Two new functions in the `Lazy.Tree` called `isInTree` and `isInForest`, and
+- a wrapper function in the `Lazy.Tree.Zipper` called `isInZipper`.
```
turboMaCk, line 36: @jhrcek this the explanation for your question BTW.

NOTES.md
```diff
@@ -0,0 +1,66 @@
...
+
+References for this step:
+
+- [Wikipedia - Mutual recursion](https://en.wikipedia.org/wiki/Mutual_recursion)
+- [Simply Scheme - Trees](https://people.eecs.berkeley.edu/~bh/pdf/ssch18.pdf) pp. 310 - 313
```
turboMaCk, line 41: Nice to see these references - I've learned a few new things myself.

NOTES.md
```diff
@@ -0,0 +1,66 @@
...
+
+- The browser can be extracted into its own reusable module.
+
+- All the non-hierarchical filter types can be extracted into their own reusable modules.
+
+- Adding new filters can be implemented with the same pattern.
+
+- Whenever a user sees a browser+filters widget on the site, they know that it works consistently.
+
+- The users can organize their data into folders as they want, because the folders are neutral.
```
turboMaCk, line 66: Thanks for writing this. It clarified a lot of details.

SECOND_STEP.md
```diff
@@ -22,15 +22,21 @@ Categories are:
...
-- [ ] In case of shared audiences ignore tree structure and display audiences in single level without any folder.
+- [x] Every-time user changes category she sees its root level.
+- [x] In case of shared audiences ignore tree structure and display audiences in single level without any folder.
 - [ ] Final solution should be compilable to html app or should contain HTML and source compilable to JavaScript included in HTML.
 
+#### Proposal
+
+My proposal is to replace the two checked requirements with the following one:
+
+- [ ] Every-time user changes category she sees the audiences of the selected categories.
```
turboMaCk, line 38: definitely possible and for some workflow even better. It was not part of the original requirements but we must give you that it made the whole problem only more interesting :+1:

public/index.html
```diff
@@ -0,0 +1,24 @@
...
+
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta name="description" content="annaghi/challenge-elm" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+
+    <title>annaghi | challenge-elm</title>
+
+    <link href="https://fonts.googleapis.com/css?family=Lato:300,400" type="text/css" rel="stylesheet" />
```
turboMaCk, line 11: attention to detail :)

src/Data/Audience.elm
turboMaCk: I agree with both.

1. Even hints [mentioned applicative style decoding](https://github.com/GlobalWebIndex/challenge-elm/blob/master/FIRST_STEP.md#hints)
2. It's still not a hard requirement and it's possible to solve it with `mapN`.

I would personally prefer seeing an applicative style here but avoiding it is not a big deal.

src/Data/Audience.elm
turboMaCk: Agree it might be clearer to do it this way.

src/Data/Audience.elm
turboMaCk: spoiler alert.

Agree that solution with `Maybe Int` would be much clearer about the special case. This seems a bit JSy.

In fact, this is isomorphic to `Maybe Int` but without utilizing type-system. On a value level, this function produces values from zero to length - 1 + extra value of -1. Just this -1 is not captured in codomain by type system using `Nothing` explicitly.

Using `Maybe` would be a better choice here but on the bright side, it's easy enough to catch and change those things in code-review.

src/Main.elm
turboMaCk: I agree with @Janiczek 's comment.

src/Main.elm
```diff
@@ -1,10 +1,405 @@
...
+
+
+withFilter : Model -> Model
+withFilter model =
+    { model | filter = Array.fromList Data.Audience.initialAudienceType }
+
+
+withErrors : String -> Model -> Model
+withErrors errorCode model =
+    { model | errors = errorCode :: model.errors }
```
turboMaCk, line 132: nice that you're collecting all errors.

turboMaCk (commented):
Nice solution!



I want to specifically highlight 3 things:

1. I believe this is the first time when someone decided to incorporate the second part which makes things even slightly more interesting :clap: 
2. It's probably the best looking version ever which attention details which go to UI and UX
3. Documentation and especially notes were as well full of details and highlight the attention to detail.

The biggest weakness in the code I see is the usage of `-1` special value and maybe the choice of product type using tuple containing 2 Bools which don't tell much of a story. Both are simple to improve though.

Now about the important part @annaghi. I've talked with my managers and we're interested in talking to you about a job position. We're definitely interested from our side but of course, it depends if you're interested in talking to us as well. In case you're not interested it's fine with us as well but in case you're at least keen to hear our offer please contact me on email marek@globalwebindex.com.

src/Lazy/Tree.elm
Janiczek: Btw @annaghi do you have an idea when would this blow up the stack? Would we eg. have to have a tree [`MAX_FN_CALLS`](http://jsfiddle.net/josh3736/Z5pLM/) deep?

(I'm even fine with no answer. Just wondering about the potential limits of this recursion)

annaghi:
Thank you @turboMaCk @Janiczek @jhrcek @kraklin and @janjelinek for being open-minded to my solution, and for your sharp eyes on details. I am going to resolve the highlighted issues within a couple of days. And yes, I have enjoyed this challenge! :sparkles:

PR #15, opened 2019-07-04
Ovechkin Anton - Challenge

owanturist:
Hello there!

I've done the code challenge.
I'm waiting for your feedback :)
And yes, the goal is to be a part of your team.

### Build

```bash
# I suppose you have elm installed
elm make src/Main.elm --optimize
```

### Things which are not done

I noticed a couple of features which are in the screenshot but not in the steps requirements, such as:
- [ ] search (top of the screenshot)
- [ ] `Audience` editing
- [ ] icons (folders, editing, filters)
- [ ] D&D for dragzone (bottom of the screenshot)
- [ ] some counter near folder's name
- [ ] ordering of the entities

I've decided to skip the points for making the bunch of changes as small as possible to make review easier. But I know how to implement all of them so it will be done if you want to.

### DIFF
--- a/src/Data/Api.elm
+++ b/src/Data/Api.elm
@@ -1,0 +1,39 @@
+module Data.Api exposing
+    ( getListOfAudienceFolders
+    , getListOfAudiences
+    )
+
+import Data.Audience as Audience exposing (Audience)
+import Data.AudienceFolder as AudienceFolder exposing (AudienceFolder)
+import Http
+import Json.Decode as Decode exposing (decodeString)
+import Process
+import Task exposing (Task)
+
+
+getListOfAudiences : Task Http.Error (List Audience)
+getListOfAudiences =
+    Task.andThen
+        (\_ ->
+            case decodeString (Decode.list Audience.decoder) Audience.audiencesJSON of
+                Err error ->
+                    Task.fail (Http.BadBody (Decode.errorToString error))
+
+                Ok audiences ->
+                    Task.succeed audiences
+        )
+        (Process.sleep 1000)
+
+
+getListOfAudienceFolders : Task Http.Error (List AudienceFolder)
+getListOfAudienceFolders =
+    Task.andThen
+        (\_ ->
+            case decodeString (Decode.list AudienceFolder.decoder) AudienceFolder.audienceFoldersJSON of
+                Err error ->
+                    Task.fail (Http.BadBody (Decode.errorToString error))
+
+                Ok audiences ->
+                    Task.succeed audiences
+        )
+        (Process.sleep 500)
--- a/src/Data/Api.elm
+++ b/src/Data/Api.elm
@@ -15,7 +15,11 @@ getListOfAudiences : Task Http.Error (List Audience)
 getListOfAudiences =
     Task.andThen
         (\_ ->
-            case decodeString (Decode.list Audience.decoder) Audience.audiencesJSON of
+            case
+                decodeString
+                    (Decode.field "data" (Decode.list Audience.decoder))
+                    Audience.audiencesJSON
+            of
                 Err error ->
                     Task.fail (Http.BadBody (Decode.errorToString error))
 
@@ -29,7 +33,11 @@ getListOfAudienceFolders : Task Http.Error (List AudienceFolder)
 getListOfAudienceFolders =
     Task.andThen
         (\_ ->
-            case decodeString (Decode.list AudienceFolder.decoder) AudienceFolder.audienceFoldersJSON of
+            case
+                decodeString
+                    (Decode.field "data" (Decode.list AudienceFolder.decoder))
+                    AudienceFolder.audienceFoldersJSON
+            of
                 Err error ->
                     Task.fail (Http.BadBody (Decode.errorToString error))
 
--- a/src/Data/Audience.elm
+++ b/src/Data/Audience.elm
@@ -34,6 +34,26 @@ type AudienceType
     | Curated
 
 
+audienceTypeDecoder : Decoder AudienceType
+audienceTypeDecoder =
+    Decode.andThen
+        (\str ->
+            case str of
+                "user" ->
+                    Decode.succeed Authored
+
+                "shared" ->
+                    Decode.succeed Shared
+
+                "curated" ->
+                    Decode.succeed Curated
+
+                unknown ->
+                    Decode.fail ("Unknown AudienceType `" ++ unknown ++ "`")
+        )
+        Decode.string
+
+
 {-| Basic type of Audience record
 -}
 type alias Audience =
@@ -49,10 +69,10 @@ type alias Audience =
 decoder : Decoder Audience
 decoder =
     Decode.map4 Audience
-        Decode.int
-        Decode.string
-        (Decode.succeed Authored)
-        (Decode.nullable Decode.int)
+        (Decode.field "id" Decode.int)
+        (Decode.field "name" Decode.string)
+        (Decode.field "type" audienceTypeDecoder)
+        (Decode.field "folder" (Decode.nullable Decode.int))
 
 
 
--- a/src/Data/Audience.elm
+++ b/src/Data/Audience.elm
@@ -1,5 +1,6 @@
 module Data.Audience exposing
     ( AudienceType(..), Audience
+    , decoder
     , audiencesJSON
     )
 
@@ -12,9 +13,15 @@ This module implements everything related to audience resource.
 
 @docs AudienceType, Audience
 
+@docs decoder
+
 @docs audiencesJSON
 
 -}
+
+import Json.Decode as Decode exposing (Decoder)
+
+
 
 -- Type definition
 
@@ -35,6 +42,17 @@ type alias Audience =
     , type_ : AudienceType
     , folder : Maybe Int
     }
+
+
+{-| Decoder of Audience record
+-}
+decoder : Decoder Audience
+decoder =
+    Decode.map4 Audience
+        Decode.int
+        Decode.string
+        (Decode.succeed Authored)
+        (Decode.nullable Decode.int)
 
 
 
--- a/src/Data/Audience.elm
+++ b/src/Data/Audience.elm
@@ -10,7 +10,9 @@ This module implements everything related to audience resource.
 
 # Interface
 
-@docs AudienceType, Audience, audienceJSON
+@docs AudienceType, Audience
+
+@docs audiencesJSON
 
 -}
 
--- a/src/Data/Store.elm
+++ b/src/Data/Store.elm
@@ -1,4 +1,12 @@
-module Data.Store exposing (AudienceFolderID, AudienceID, Store)
+module Data.Store exposing
+    ( AudienceFolderID
+    , AudienceID
+    , AudienceLevel
+    , Store
+    , getLevelByFolderID
+    , getRootLevel
+    , init
+    )
 
 import Data.Audience exposing (Audience)
 import Data.AudienceFolder exposing (AudienceFolder)
@@ -15,31 +23,37 @@ type alias AudienceID =
 
 
 type alias AudienceLevel =
+    { folders : List AudienceFolder
+    , audiences : List Audience
+    }
+
+
+type alias Level =
     { folders : Set AudienceFolderID
     , audiences : Set AudienceID
     }
 
 
-emptyLevel : AudienceLevel
+emptyLevel : Level
 emptyLevel =
-    AudienceLevel Set.empty Set.empty
+    Level Set.empty Set.empty
 
 
-insertFolderToLevel : AudienceFolderID -> AudienceLevel -> AudienceLevel
+insertFolderToLevel : AudienceFolderID -> Level -> Level
 insertFolderToLevel folderID level =
     { level | folders = Set.insert folderID level.folders }
 
 
-insertAudienceToLevel : AudienceID -> AudienceLevel -> AudienceLevel
+insertAudienceToLevel : AudienceID -> Level -> Level
 insertAudienceToLevel audienceID level =
     { level | audiences = Set.insert audienceID level.audiences }
 
 
 type alias Levels =
-    Dict AudienceFolderID AudienceLevel
+    Dict AudienceFolderID Level
 
 
-updateOnLevels : AudienceFolderID -> (AudienceLevel -> AudienceLevel) -> Levels -> Levels
+updateOnLevels : AudienceFolderID -> (Level -> Level) -> Levels -> Levels
 updateOnLevels folderID tagger levels =
     Dict.update folderID (Just << tagger << Maybe.withDefault emptyLevel) levels
 
@@ -49,7 +63,7 @@ type Store
         { folders : Dict AudienceFolderID AudienceFolder
         , audiences : Dict AudienceID Audience
         }
-        { root : AudienceLevel
+        { root : Level
         , levels : Levels
         }
 
@@ -100,18 +114,43 @@ init listOfFolders listOfAudiences =
     Store data2 relations2
 
 
+extractListFromData : Set comparable -> Dict comparable { a | id : comparable } -> List { a | id : comparable }
+extractListFromData ids entities =
+    Set.foldr
+        (\id acc ->
+            case Dict.get id entities of
+                Nothing ->
+                    acc
 
--- getLevelByFolderID : Maybe AudienceFolderID -> Store -> AudienceLevel
--- getLevelByFolderID targetFolderID (Store { folders, audiences }) =
---     { folders =
---         folders
---             |> Dict.filter (\_ folder -> targetFolderID == folder.parent)
---             |> Dict.values
---     , audiences =
---         audiences
---             |> Dict.filter (\_ audience -> targetFolderID == audience.folder)
---             |> Dict.values
---     }
--- getRoot : Store -> AudienceLevel
--- getRoot =
---     getLevelByFolderID Nothing
+                Just entity ->
+                    entity :: acc
+        )
+        []
+        ids
+
+
+extractLevelByParentID : Maybe AudienceFolderID -> Store -> Maybe AudienceLevel
+extractLevelByParentID mFolderID (Store data { root, levels }) =
+    Maybe.map
+        (\{ folders, audiences } ->
+            { folders = extractListFromData folders data.folders
+            , audiences = extractListFromData audiences data.audiences
+            }
+        )
+        (case mFolderID of
+            Nothing ->
+                Just root
+
+            Just folderID ->
+                Dict.get folderID levels
+        )
+
+
+getRootLevel : Store -> Maybe AudienceLevel
+getRootLevel =
+    extractLevelByParentID Nothing
+
+
+getLevelByFolderID : AudienceFolderID -> Store -> Maybe AudienceLevel
+getLevelByFolderID =
+    extractLevelByParentID << Just
--- a/src/Data/Store.elm
+++ b/src/Data/Store.elm
@@ -5,6 +5,7 @@ module Data.Store exposing
     , Selection(..)
     , Selector(..)
     , Store
+    , getFolderByID
     , init
     , select
     )
@@ -149,8 +150,10 @@ extractListFromData ids entities =
 
 type Selector
     = OnlyShared
-    | OnlyCuratedIn (Maybe AudienceFolderID)
-    | OnlyAuthoredIn (Maybe AudienceFolderID)
+    | OnlyCurated
+    | OnlyCuratedIn AudienceFolderID
+    | OnlyAuthored
+    | OnlyAuthoredIn AudienceFolderID
 
 
 type Selection
@@ -168,13 +171,13 @@ select selector (Store data { root, levels, shared }) =
                 (extractListFromData shared data.audiences)
                 |> Root
 
-        OnlyCuratedIn Nothing ->
+        OnlyCurated ->
             AudienceLevel
                 (extractListFromData root.folders data.folders)
                 (extractListFromData root.curated data.audiences)
                 |> Root
 
-        OnlyCuratedIn (Just folderID) ->
+        OnlyCuratedIn folderID ->
             case ( Dict.get folderID data.folders, Dict.get folderID levels ) of
                 ( Just folder, Just level ) ->
                     AudienceLevel
@@ -185,13 +188,13 @@ select selector (Store data { root, levels, shared }) =
                 _ ->
                     NotFound folderID
 
-        OnlyAuthoredIn Nothing ->
+        OnlyAuthored ->
             AudienceLevel
                 (extractListFromData root.folders data.folders)
                 (extractListFromData root.authored data.audiences)
                 |> Root
 
-        OnlyAuthoredIn (Just folderID) ->
+        OnlyAuthoredIn folderID ->
             case ( Dict.get folderID data.folders, Dict.get folderID levels ) of
                 ( Just folder, Just level ) ->
                     AudienceLevel
@@ -201,3 +204,8 @@ select selector (Store data { root, levels, shared }) =
 
                 _ ->
                     NotFound folderID
+
+
+getFolderByID : AudienceFolderID -> Store -> Maybe AudienceFolder
+getFolderByID folderID (Store data _) =
+    Dict.get folderID data.folders
--- a/src/Data/Store.elm
+++ b/src/Data/Store.elm
@@ -4,6 +4,7 @@ module Data.Store exposing
     , AudienceLevel
     , Store
     , getLevelByFolderID
+    , getParentFolderID
     , getRootLevel
     , init
     )
@@ -84,13 +85,28 @@ init listOfFolders listOfAudiences =
         ( data1, relations1 ) =
             List.foldr
                 (\folder ( data, relations ) ->
+                    let
+                        nextLevels =
+                            Dict.update folder.id
+                                (\level ->
+                                    case level of
+                                        Nothing ->
+                                            Just emptyLevel
+
+                                        _ ->
+                                            level
+                                )
+                                relations.levels
+                    in
                     ( { data | folders = Dict.insert folder.id folder data.folders }
                     , case folder.parent of
                         Nothing ->
-                            { relations | root = insertFolderToLevel folder.id relations.root }
+                            { root = insertFolderToLevel folder.id relations.root
+                            , levels = nextLevels
+                            }
 
                         Just parentID ->
-                            { relations | levels = updateOnLevels parentID (insertFolderToLevel folder.id) relations.levels }
+                            { relations | levels = updateOnLevels parentID (insertFolderToLevel folder.id) nextLevels }
                     )
                 )
                 ( data0, relations0 )
@@ -129,28 +145,24 @@ extractListFromData ids entities =
         ids
 
 
-extractLevelByParentID : Maybe AudienceFolderID -> Store -> Maybe AudienceLevel
-extractLevelByParentID mFolderID (Store data { root, levels }) =
+getRootLevel : Store -> AudienceLevel
+getRootLevel (Store data { root }) =
+    { folders = extractListFromData root.folders data.folders
+    , audiences = extractListFromData root.audiences data.audiences
+    }
+
+
+getLevelByFolderID : AudienceFolderID -> Store -> Maybe AudienceLevel
+getLevelByFolderID folderID (Store data { levels }) =
     Maybe.map
         (\{ folders, audiences } ->
             { folders = extractListFromData folders data.folders
             , audiences = extractListFromData audiences data.audiences
             }
         )
-        (case mFolderID of
-            Nothing ->
-                Just root
-
-            Just folderID ->
-                Dict.get folderID levels
-        )
+        (Dict.get folderID levels)
 
 
-getRootLevel : Store -> Maybe AudienceLevel
-getRootLevel =
-    extractLevelByParentID Nothing
-
-
-getLevelByFolderID : AudienceFolderID -> Store -> Maybe AudienceLevel
-getLevelByFolderID =
-    extractLevelByParentID << Just
+getParentFolderID : AudienceFolderID -> Store -> Maybe AudienceFolderID
+getParentFolderID folderID (Store data _) =
+    Maybe.andThen .parent (Dict.get folderID data.folders)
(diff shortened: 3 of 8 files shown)


owanturist:
READY

turboMaCk:
Hi, @owanturist thanks for the PR and interest. We will review it ASAP. Anyway, there might be a complication with you being located in Russia as our current policy is that all contractors must be entities under Europian Union laws, but I'm going to check with managers and legal department to see how hard rule it is.

src/Data/Api.elm
```diff
@@ -0,0 +1,47 @@
...
+                    (Decode.field "data" (Decode.list Audience.decoder))
+                    Audience.audiencesJSON
+            of
+                Err error ->
+                    Task.fail (Http.BadBody (Decode.errorToString error))
+
+                Ok audiences ->
+                    Task.succeed audiences
+        )
+        (Process.sleep 1000)
```
Janiczek, line 29: Are these `Process.sleep`s there to test UI loading/... states?

src/Main.elm
```diff
@@ -1,10 +1,418 @@
...
+
+viewFilterButton : List (Element.Attribute msg) -> Element msg -> Element msg
+viewFilterButton attributes label =
+    button
+        (Element.width Element.fill
+            :: Element.paddingXY 6 18
+            :: Element.Font.center
+            :: attributes
+        )
+        { onPress = Nothing
```
Janiczek, line 256: Are there some bugs or problems when using `onPress`? I see you're passing `onClick` in `attributes` instead - but I'd expect the events to be passed into `onPress`...

src/Main.elm
```diff
@@ -1,10 +1,418 @@
...
+    = LoadAudienceFoldersDone (Result Http.Error (List AudienceFolder))
+    | LoadAudiencesDone (Result Http.Error (List Audience))
+    | GoDown AudienceFolderID
+    | GoUp
+    | SetFilterBy AudienceType
+
+
+update : Msg -> Model -> ( Model, Cmd Msg )
+update msg model =
+    case ( msg, model ) of
```
Janiczek, line 111: I'd prefer having multiple `update` functions - one for each `Model` case ... less stuff to think about

src/Data/Store.elm
```diff
@@ -0,0 +1,211 @@
...
+            , audiences = Dict.empty
+            }
+
+        relations0 =
+            { root = emptyLevel
+            , levels = Dict.fromList (List.map (\folder -> ( folder.id, emptyLevel )) listOfFolders)
+            , shared = []
+            }
+
+        ( data1, relations1 ) =
```
Janiczek: I'd prefer these `(data1, relations1)` and `(data2, relations2)` things to use some extracted function whose name tells us better what actually happened (ie. something like `processFolders` and `processAudiences`)

Janiczek:
As we're using Zippers in our real codebase, it was interesting seeing a different approach - your `Data.Store`. I can't decide if `Selector` wasn't a bit overengineered, but the overall `Dict AudienceFolderId ...` approach seems fine to me :slightly_smiling_face: Nice job!

src/Main.elm
```diff
@@ -1,10 +1,418 @@
...
+                (List.range 0 2)
+            )
+        ]
+
+
+viewFailed : Http.Error -> Element msg
+viewFailed err =
+    case err of
+        Http.BadUrl url ->
+            text ("Wrong request url is provided: `" ++ url ++ "`")
```
jhrcek, line 360: nitpick: text could be extracted out of each case branch

src/Main.elm
```diff
@@ -1,10 +1,418 @@
...
+    [ Element.width Element.fill
+    , Element.padding 20
+    ]
+
+
+viewFolder : AudienceFolder -> Element Msg
+viewFolder folder =
+    button
+        (Element.Background.color (Element.rgb255 26 113 153)
+            :: Element.Font.color (Element.rgb255 255 255 255)
```
jhrcek, line 192: It might make sense to extract these color attributes to constants, e.g. `black = Element.rgb255 255 255 255`

src/Data/Api.elm
owanturist: yes, that's right
there is a UI skeleton so I'd like you to see that

src/Main.elm
```diff
@@ -1,10 +1,418 @@
...
+    | GoDown AudienceFolderID
+    | GoUp
+    | SetFilterBy AudienceType
+
+
+update : Msg -> Model -> ( Model, Cmd Msg )
+update msg model =
+    case ( msg, model ) of
+        ( LoadAudienceFoldersDone (Err err), _ ) ->
+            ( Failed err, Cmd.none )
```
jhrcek, line 113: All branches have `Cmd.none` repeating. It might make sense to have pure update helper 
`updatePure : Msg -> Model -> Model` and then call it from update:
`update msg model = ( updatePure msg model, Cmd.none )`

src/Main.elm
owanturist: Yes, that's right. I find a bit strange the way of `button` arguments like
```elm
{ onPress : Maybe msg
, label : Element msg
}
```

I tested different combinations passing `onPress` and `onClick` and the `onClick` has higher priority.

src/Main.elm
owanturist: That `update` is too small to be divided I think.

owanturist:
@Janiczek the `Selector` was created to follow the rule "Make impossible state impossible", so that's why it there :)

src/Data/Store.elm
owanturist: I've dave a little bit deeper and remembered
why I'm using the composite (`case (msg, model of`) approach.

Let's say we have `Model` and `Msg`.
Each of the variants must be handled according postfix (`ModelA` with `MsgA`, etc...).

```elm
type Model
    = ModelA Int
    | ModelB Int
    | ModelC Int


type Msg
    = MsgA Int
    | MsgB Int
    | MsgC Int
```

The composite function will look like this:

```elm
update : Msg -> Model -> Model
update msg model =
    case ( msg, model ) of
        ( MsgA a, ModelA x ) ->
            ModelB (a + x)

        ( MsgA _, _ ) ->
            model

        ( MsgB b, ModelB x ) ->
            ModelC (b * x)

        ( MsgB _, _ ) ->
            model

        ( MsgC c, ModelC x ) ->
            ModelA (c - x)

        ( MsgC _, _ ) ->
            model
```

It prevents you from not handling new variants.
Let's say we add `MsgAA` to handle when `ModelA`:


```elm
type Msg
    = MsgA Int
    | MsgAA
    | MsgB Int
    | MsgC Int

-- compiling error
```

Compiler will say us to handle the case:

```elm
update : Msg -> Model -> Model
update msg model =
    case ( msg, model ) of
        ( MsgA a, ModelA x ) ->
            ModelB (a + x)

        ( MsgA _, _ ) ->
            model

        ( MsgAA, ModelA x ) ->
            ModelB x

        ( MsgAA _, _ ) ->
            model

        ( MsgB b, ModelB x ) ->
            ModelC (b * x)

        ( MsgB _, _ ) ->
            model

        ( MsgC c, ModelC x ) ->
            ModelA (c - x)

        ( MsgC _, _ ) ->
            model
```

But let's separate the `update` to multiple functions:

```elm
updateA : Msg -> Int -> Model
updateA msg x =
    case msg of
        MsgA a ->
            ModelB (a + x)

        _ ->
            ModelA x


updateB : Msg -> Int -> Model
updateB msg x =
    case msg of
        MsgB b ->
            ModelC (b * x)

        _ ->
            ModelB x


updateC : Msg -> Int -> Model
updateC msg x =
    case msg of
        MsgC c ->
            ModelA (c - x)

        _ ->
            ModelC x
```


The problem is that when I add one more Msg for ModelA
the compiler will say me nothing but the case will not be handled.

```elm
type Msg
    = MsgA Int
    | MsgAA
    | MsgB Int
    | MsgC Int

-- no more changes but still compiling
```

To prevent this situation I could handle all of the cases dirrectly:

```elm
updateA : Msg -> Int -> Model
updateA msg x =
    case msg of
        MsgA a ->
            ModelB (a + x)

        MsgB _ ->
            ModelA x

        MsgC _ ->
            ModelA x


updateB : Msg -> Int -> Model
updateB msg x =
    case msg of
        MsgA _ ->
            ModelB x

        MsgB b ->
            ModelC (b * x)

        MsgC _ ->
            ModelB x


updateC : Msg -> Int -> Model
updateC msg x =
    case msg of
        MsgA _ ->
            ModelC x

        MsgB _ ->
            ModelC x

        MsgC c ->
            ModelA (c - x)
```

As you can see the previous problem is solved, but another created:
you mast describe too much.

Otherwise composite approach prevent you from both
of the issues. `update` is becoming bigger and bigger but instead of
splitting by model you could split by cases, like:


```elm
hugeACase : Int -> Int -> Model
hugeACase a x =
    ModelB (a + x)


hugeBCase : Int -> Int -> Model
hugeBCase b x =
    ModelC (b * x)


hugeCCase : Int -> Int -> Model
hugeCCase c x =
    ModelC (c - x)


update : Msg -> Model -> Model
update msg model =
    case ( msg, model ) of
        ( MsgA a, ModelA x ) ->
            hugeACase a x

        ( MsgA _, _ ) ->
            model

        ( MsgB b, ModelB x ) ->
            hugeBCase b x

        ( MsgB _, _ ) ->
            model

        ( MsgC c, ModelC x ) ->
            hugeCCase

        ( MsgC _, _ ) ->
            model
```

src/Main.elm
owanturist: I've dave a little bit deeper and remembered why I'm using the composite `case (msg, model) of` approach.

Let's say we have `Model` and `Msg`. Each of the variants must be handled according postfix (`ModelA` with `MsgA`, etc...).

```elm
type Model
    = ModelA Int
    | ModelB Int
    | ModelC Int


type Msg
    = MsgA Int
    | MsgB Int
    | MsgC Int
```

The composite function will look like this:

```elm
update : Msg -> Model -> Model
update msg model =
    case ( msg, model ) of
        ( MsgA a, ModelA x ) ->
            ModelB (a + x)

        ( MsgA _, _ ) ->
            model

        ( MsgB b, ModelB x ) ->
            ModelC (b * x)

        ( MsgB _, _ ) ->
            model

        ( MsgC c, ModelC x ) ->
            ModelA (c - x)

        ( MsgC _, _ ) ->
            model
```

It prevents you from not handling new variants. Let's say we add `MsgAA` to handle when `ModelA`:


```elm
type Msg
    = MsgA Int
    | MsgAA
    | MsgB Int
    | MsgC Int

-- compiling error
```

Compiler will say us to handle the case:

```elm
update : Msg -> Model -> Model
update msg model =
    case ( msg, model ) of
        ( MsgA a, ModelA x ) ->
            ModelB (a + x)

        ( MsgA _, _ ) ->
            model

        ( MsgAA, ModelA x ) ->
            ModelB x

        ( MsgAA _, _ ) ->
            model

        ( MsgB b, ModelB x ) ->
            ModelC (b * x)

        ( MsgB _, _ ) ->
            model

        ( MsgC c, ModelC x ) ->
            ModelA (c - x)

        ( MsgC _, _ ) ->
            model
```

But let's separate the `update` to multiple functions:

```elm
updateA : Msg -> Int -> Model
updateA msg x =
    case msg of
        MsgA a ->
            ModelB (a + x)

        _ ->
            ModelA x


updateB : Msg -> Int -> Model
updateB msg x =
    case msg of
        MsgB b ->
            ModelC (b * x)

        _ ->
            ModelB x


updateC : Msg -> Int -> Model
updateC msg x =
    case msg of
        MsgC c ->
            ModelA (c - x)

        _ ->
            ModelC x
```


The problem is that when I add one more Msg for ModelA the compiler will say me nothing but the case will not be handled.

```elm
type Msg
    = MsgA Int
    | MsgAA
    | MsgB Int
    | MsgC Int

-- no more changes but still compiling
```

To prevent this situation I could handle all of the cases directly:

```elm
updateA : Msg -> Int -> Model
updateA msg x =
    case msg of
        MsgA a ->
            ModelB (a + x)

        MsgB _ ->
            ModelA x

        MsgC _ ->
            ModelA x


updateB : Msg -> Int -> Model
updateB msg x =
    case msg of
        MsgA _ ->
            ModelB x

        MsgB b ->
            ModelC (b * x)

        MsgC _ ->
            ModelB x


updateC : Msg -> Int -> Model
updateC msg x =
    case msg of
        MsgA _ ->
            ModelC x

        MsgB _ ->
            ModelC x

        MsgC c ->
            ModelA (c - x)
```

As you can see the previous problem is solved, but another created: you mast describe too much.

Otherwise composite approach prevent you from both of the issues. `update` is becoming bigger and bigger but instead of splitting by model you could split by cases, like:


```elm
hugeACase : Int -> Int -> Model
hugeACase a x =
    ModelB (a + x)


hugeBCase : Int -> Int -> Model
hugeBCase b x =
    ModelC (b * x)


hugeCCase : Int -> Int -> Model
hugeCCase c x =
    ModelC (c - x)


update : Msg -> Model -> Model
update msg model =
    case ( msg, model ) of
        ( MsgA a, ModelA x ) ->
            hugeACase a x

        ( MsgA _, _ ) ->
            model

        ( MsgB b, ModelB x ) ->
            hugeBCase b x

        ( MsgB _, _ ) ->
            model

        ( MsgC c, ModelC x ) ->
            hugeCCase

        ( MsgC _, _ ) ->
            model
```

owanturist:
@turboMaCk I had a talk with Roman Salnikov (@bardt) and he told me that it's possible to work with you from Russia :) hope nothing changed from that moment

src/Main.elm
Janiczek: Basically for me this boils into "in one big update you can go by Msgs first, Models second - you handle all Msgs and it doesn't matter if you wildcard Models" ... compared to "separate updates for each Model would have to handle all Msgs too - more boilerplate" ... Fair enough!

src/Main.elm
jhrcek: I don't think there's a bug in elm-ui. onPress works fine for me. Maybe it was confusing why onPress message is wrapped in Maybe? AFAIK that was a design decision of the author with Maybe representing disabled button (to express that onPress on disabled button can't produce any messages)

src/Data/Store.elm
```diff
@@ -0,0 +1,233 @@
...
+
+extractListFromData : List comparable -> Dict comparable { a | id : comparable } -> List { a | id : comparable }
+extractListFromData ids entities =
+    List.filterMap (\id -> Dict.get id entities) ids
+
+
+type Selector
+    = OnlyShared
+    | OnlyCurated
+    | OnlyCuratedIn AudienceFolderID
```
jhrcek, line 176: I slightly don't like how the "what we're displaying" (shared/curated/authored) is coupled to "where we're in the tree" (root/non-root folder). But I don't have any specific suggestions how to do this differently with this design..

jhrcek (commented):
Nice solutions. I like that you thought about initialization of the model and made a separate view function for it. Only couple of nitpicks inline.

src/Main.elm
```diff
@@ -1,10 +1,418 @@
...
+            , Cmd.none
+            )
+
+        ( LoadAudienceFoldersDone _, _ ) ->
+            ( model, Cmd.none )
+
+        ( LoadAudiencesDone (Err err), _ ) ->
+            ( Failed err, Cmd.none )
+
+        ( LoadAudiencesDone (Ok listOfAudiences), Initialising state ) ->
```
janjelinek, line 126: To be honest this whole approach seems to me too complicated and hard to understand what is covered and what is not. Yes, you have a compiler to say you, what case has to be covered, but in this line is an example of a false sense of security. 
Image when you will have the ability to reload audiences/folders even if they are already loaded. 
Someone who did not write this code will use `Api.getListOfAudiences` together with existing Msg `LoadAudiencesDone` and WTF moment is here. You can say, it's no big issue, but when you have huge codebase full of those hidden non-obvious logic, it's always pain.

src/Main.elm
```diff
@@ -1,10 +1,418 @@
...
+viewGoUp : AudienceFolder -> Element Msg
+viewGoUp parentFolder =
+    button
+        (Element.Border.solid
+            :: Element.Border.width 2
+            :: Element.Border.color (Element.rgb255 26 113 153)
+            :: Element.Font.color (Element.rgb255 26 113 153)
+            :: stylesButton
+        )
+        { onPress = Just GoUp
```
janjelinek, line 219: When you add destination folder to `GoUp` Msg you will have better UX for cases when browser froze and user accidentally do more clicks. In your cases if there will be multiple messages emitted from the same level you will end up in root folder. But if you will have `GoUp parentFolder` and miltiple messages will be emitted, all are producing same output (go one level up).
Also with that approach, you can simplify your `update` `case` because you can have only one message for both actions like `GoToFolder (Maybe AudienceFolderID)`.

Oh wait, I see now, it requires another argument here because `parentFolder` int this case is not `parentFolder` but `curentFolder` 🤔

src/Main.elm
kraklin: @jhrcek - I know, it is my OCD, but I can't resist. Take it as really small nitpick to your nitpick and you can completly ignore this comment: `Element.rgb 255 255 255` would be `white` not `black` :D

src/Data/Api.elm
```diff
@@ -0,0 +1,47 @@
...
+getListOfAudienceFolders : Task Http.Error (List AudienceFolder)
+getListOfAudienceFolders =
+    Task.andThen
+        (\_ ->
+            case
+                decodeString
+                    (Decode.field "data" (Decode.list AudienceFolder.decoder))
+                    AudienceFolder.audienceFoldersJSON
+            of
+                Err error ->
```
kraklin, line 41: I'd prefer `Result.map` and `Result.withDefault` instead of `case`. Also in both functions the handling of result of `decodeString` is the same so you can create something like: 

```
toTask : Result x a -> Task Http.Error a
toTask result =
    Result.map Task.succeed result
    |> Result.withDefault (Task.fail (Http.BadBody (Decode.errorToString error)))
```

Then you'll have the responsibilities clearly separated one from another

kraklin (commented):
Just small nitpicks. I do agree with others on the unnecessary complexity of the `case` in `update` function though.

src/Data/Api.elm
```diff
@@ -0,0 +1,47 @@
...
+                    Task.fail (Http.BadBody (Decode.errorToString error))
+
+                Ok audiences ->
+                    Task.succeed audiences
+        )
+        (Process.sleep 500)
```
turboMaCk, line 47: you could have probably created abstraction for simulating the sleep eventhough you use different  values in both cases.

src/Data/Api.elm
```diff
@@ -0,0 +1,47 @@
...
+    ( getListOfAudienceFolders
+    , getListOfAudiences
+    )
+
+import Data.Audience as Audience exposing (Audience)
+import Data.AudienceFolder as AudienceFolder exposing (AudienceFolder)
+import Http
+import Json.Decode as Decode exposing (decodeString)
+import Process
+import Task exposing (Task)
```
turboMaCk, line 11: I like those imports (exposing types and single word qualification). this is what we do as well.

src/Data/Audience.elm
```diff
@@ -35,6 +64,17 @@ type alias Audience =
...
 
+{-| Decoder of Audience record
+-}
+decoder : Decoder Audience
+decoder =
+    Decode.map4 Audience
+        (Decode.field "id" Decode.int)
+        (Decode.field "name" Decode.string)
+        (Decode.field "type" audienceTypeDecoder)
+        (Decode.field "folder" (Decode.nullable Decode.int))
```
turboMaCk, line 75: applicative style is usually simpler to extend though there is nothing wrong with using `map4`.

src/Data/Store.elm
turboMaCk: agree. not a big deal though

src/Main.elm
```diff
@@ -1,10 +1,418 @@
...
+
+---------------
+-- M O D E L --
+---------------
+
+
+type alias InitialisingState =
+    { folders : Maybe (List AudienceFolder)
+    , audiences : Maybe (List Audience)
+    }
```
turboMaCk, line 29: this type imo shows one of the weaknesses of tea.. things like this has to be dealth with using custom state handling. Aff monad would be probably much nicer for modeling this :/

src/Main.elm
```diff
@@ -1,10 +1,418 @@
...
+
+
+turnInitialising : InitialisingState -> Model
+turnInitialising state =
+    case Maybe.map2 Store.init state.folders state.audiences of
+        Nothing ->
+            Initialising state
+
+        Just store ->
+            Succeed store (SucceedState Root Audience.Authored)
```
turboMaCk, line 79: there I think avoid pattern matching would be nicer.

```elm
Maybe.map2 Store.init state.folders state.audiences
  |> Maybe.map (Basics.Extra.flip Succeed (SucceedState Root Audience.Authored))
  |> Maybe.withDefault (Initialising state)
```

though readability with pattern matching is not worse, maybe even better if one preferse to think about branches rather than composition.

src/Main.elm
turboMaCk: on the other hand this assumption might change and if so updating this implementation to include Cmd in some branch will be simpler.

src/Main.elm
```diff
@@ -1,10 +1,418 @@
...
+
+
+view : Model -> Browser.Document Msg
+view model =
+    Browser.Document "Challenge"
+        [ Element.layout
+            [ Element.height Element.fill
+            ]
+            (viewLayout model)
+        ]
```
turboMaCk, line 402: I like the fact that you're using flow of the code I prefer - function that uses function is bellow that function where possible rather than other way around which seems to be generally prefered flow. But that might be personal bias.

turboMaCk (commented):
I have to admit there are a lot of aparently small codestyle things I really like (imports, flow of code etc) which are pretty rare. I would personally use `andMap` from decode-extra for decoding (Or define own using `Decode.map2 (|>)` but it might be just me hating `mapN` for no reason. Personally I bit dislike using that much patternmatching as in my experience it's harder to maitain small functions with comprehensible api with too much branching.

Over all this is very nice solution. I'm really sorry to say that at this moment we can hire only people from EU countries though. At the time Roman S. started we had not so many rules but since then because of GDPR, internal HR departmenet and legal department and because we now have offices in 3 EU countries company policy is to hire EU residents only and preferably only those who are from countries we have office (and thus local company) though. This might be only temporary situation as company tries to primarly establish permanent offices and for that we need staff but I frankly don't see into those things that much myslf. All I can promis at this moment that we might contact you in case this changes in the future in case you'll be still interested.

If you don't mind I can also give your contact to a fried who is currently looking for elm developers for an interesting project. I just asked him and he will be definitely interested in talking to you. Please let me know;)

src/Main.elm
owanturist: @jhrcek I didn't know that `Nothing` disabled the button. But anyway, I don't find the current button's API clean.

src/Main.elm
owanturist: The app is too small to try make some colour stuff here. Like `primaryColor`, `secondary...`, etc. I mean, if I'd decide to create `black` or `white` I **must** do the same for the rest to be consistent. I just didn't want focus on it.

src/Main.elm
owanturist: Here is elm app https://app.texenergo.com with 44k lines of elm code in 109 files which include routing and handling 36 url routes. Still no problem with the approach.

src/Data/Api.elm
owanturist: The idea is good but I prefer `case..of` in purpose to escape Lisp looking code.

src/Data/Api.elm
owanturist: Something like this:

```elm
imitateLoading: Int -> Task Never ()
imitateLoading timeout =
	Process.sleep timeout
```

src/Data/Audience.elm
owanturist: I prefer to use `Decode.map*` where it's possible. I just like the `Decode`'s API.

**Evan's [note](https://package.elm-lang.org/packages/elm/json/latest/Json-Decode#map)**:
>  If you run out of map functions, take a look at [elm-json-decode-pipeline](https://package.elm-lang.org/packages/NoRedInk/elm-json-decode-pipeline/latest) which makes it easier to handle large objects, but produces lower quality type errors.

src/Data/Store.elm
owanturist: @jhrcek it's just for making impossible states impossible.

src/Main.elm
owanturist: I've gone thru the time when I tried to avoid any `case..of` but use some sort of flows and `map*` and `flip` and `curry` and `uncurry`... Then I realised that `case..of` is 10x times more clear and readable than deep nesting of parentheses. That's also what @rtfeldman told in the Elm Europe 2019 conference as well.

Using of `Basics.Extra.flip` adds one more extra knowledge which developers should know. It's easy but I'm 100% sure it doesn't help really much. It just making thinks more wrapped and requires more time to read, understand and change.

What will happen when `Succeed` get one more argument? Let's say it become `Succeed Store SucceedState Int`:

```elm
turnInitialising : Int -> InitialisingState -> Model
turnInitialising someInt state =
    Maybe.map2 Store.init state.folders state.audiences
        |> Maybe.map (flip Succeed (SucceedState Root Audience.Authored))
        |> Maybe.map ((|>) someInt)
        |> Maybe.withDefault (Initialising state)
```

or 

```elm
turnInitialising : Int -> InitialisingState -> Model
turnInitialising someInt state =
    Maybe.map2 Store.init state.folders state.audiences
        |> Maybe.map ((|>) someInt << flip Succeed (SucceedState Root Audience.Authored))
        |> Maybe.withDefault (Initialising state)
```

VS

```elm

turnInitialising : Int -> InitialisingState -> Model
turnInitialising someInt state =
    case Maybe.map2 Store.init state.folders state.audiences of
        Nothing ->
            Initialising state

        Just store ->
            Succeed store (SucceedState Root Audience.Authored) someInt
```

I bet the last piece of code is quite more readable and easy changeable.

src/Main.elm
owanturist: At the moment of implementing the logic there was not ability to load the data when the app already have it. Make impossible state impossible.

src/Main.elm
owanturist: Here `text` function is too small to be extracted in my opinion. The `Http.BadBody` case also could became something more then just `text` like:

```elm
div
    []
    [ text "Server response incorrectly"
    , code [] [ text message ]
    ]

```
The case will require almost no cost to update.

src/Main.elm
owanturist: The code is like a story..)

src/Main.elm
owanturist: I've genteelly picked the name for the messages `GoUp` and `GoDown AudienceFolderID` because they clearly show what happens and it will be handled exactly like `GoToFolder (Maybe AudienceFolderID)`:

```elm
case msg of
    GoToFolder Nothing ->
        ...
    GoToFolder (Just id) ->
        ...
```
same with
```elm
case msg of
    GoUp ->
        ...
    GoDown id ->
        ...
```

>  for cases when browser froze and user accidentally do more clicks. In your cases if there will be multiple messages emitted from the same level you will end up in root folder. But if you will have GoUp parentFolder and miltiple messages will be emitted, all are producing same output (go one level up)

It's still JS and this part is synchronous and if this case happen it will work exactly like with `GoToFolder (Maybe AudienceFolderID)`:
- UI got frozen
- user clicks to the button [1]
- user clicks again [2]
- and again [3]
- UI is unlocked
- [1] event bubbles up and single JS stream is working under calling update (state will be changed) and then render according the new state
- UI updated and [2] event bubbles and calls the `GoToFolder` with argument from the new state from ^
- similar for [3] as ^

Result will be the same.

owanturist (commented):
@turboMaCk yes, please share my contacts to the guy. I'll be glad to contribute to some interesting elm project.

src/Data/Audience.elm
turboMaCk: Like everything, this is about tradeoffs and probably even certain experience biases. For Haskell folks when it comes to this type of composition Applicative style seems like the most elegant solution to the problem usually even though for instance @jhrcek also said on some occasions that he prefers mapN.

There is another issue with using mapN in bigger teams and that is that if 2 people in 2 PRs add different attribute it produces a bit more complicated conflict.

In our team, we use json-extra because at the time there was not decode-pipeline package yet and `andMap` is more general than functions from pipeline package which makes the assumption that applicative style is used over the object.

No super-strong preferences from my side though. I think this is not that crucial thing usually. One probably favors the consistency over anything else when it comes to things like this.

The only thing this was raised several times was probably that [hints in the requirements mentions this](https://github.com/GlobalWebIndex/challenge-elm/blob/master/FIRST_STEP.md#hints)

src/Main.elm
turboMaCk: I know Richard's arguments and I don't say they have no point but there are also important contra arguments. And I don't mean the fact that pattern matching makes compiler slower, that's not even that big deal since 0.19.

Pattern matching is a good tool for dealing with branching. But in my opinion, branching should not be done everywhere in the code. More likely it's desirable to avoid branching in some parts of logic and push branching to a very low level and very high-level parts of control flow as much as possible. using pattern matching one can do almost anything in any of branches as long as the result type checks. The composition is much simpler to reason about as it must obey certain rules. Also writing code that can compose often leads to smaller and better self-contained building blocks. These composition rules are the primary tool we use in FP to reason about the program so it's required to be comfortable with them. Those patterns often let you define small functions that you can compose using the map and then build bigger blocks. With pattern matching this was of building larger structures from smaller ones is much more obfuscated.

The burden of choosing the right approach is still on a developer. But I think the argument of having to know less is something I can't buy into generally. If anything people should be encouraged to learn more.

turboMaCk:
thanks, @owanturist for your time and constructive discussion. Petr will get in touch with you.

src/Main.elm
owanturist: Agree, sometimes casing is not needed especially when you work with well known and final solutions, like `Maybe`, `Result`, etc. But in case when you use custom types in a way like here it means that the types will change often and so it's not a good place to try piping.

PR #17, opened 2020-09-30
Sébastien Besnier submission

sebsheep:
Hello!

Here is my submission!

I've focused a lot on error handling, trying to be "bullet proof" in case of inconsistencies from the backend and giving nice error messages (maybe elm influenced me a *little bit* :D ).

As discussed with @Janiczek on slack, I've added some data in the fixtures to demonstrate how this error handling works. 

With more time, I'd polish the style a bit more:
* handle scrolling more properly,
* display a nice loader (pulsing fake cards for the audience & folders),
* fix a bug with long folder names: in this case, the number of items in the folder is not visible,
* work on error style.

And I'd also have to deal with "cycling folders" (e.g. if folder 1 has 2 as parent and folder 2 has 1 as parent). For now, I think there are silently not displayed, which is not nice.

Ping me on slack!


sebsheep:
Just forgot: to run the app, launch the `./build.sh` command. It will build everything needed in the `public` folder. You just then have to serve this folder (e.g. you can use `elm reactor` for this purpose).

src/Data/Audience.elm
```diff
@@ -36,6 +40,45 @@ type alias Audience =
...
+decoder =
+    -- I know about Json.Decode.Pipeline but for small projects, I prefer to
+    -- limit the number of dependencies. Plus, using `mapX` leads to easier
+    -- to understand error messages than the applicative version.
+    Json.Decode.map4 Audience
+        (Json.Decode.field "id" Json.Decode.int)
+        (Json.Decode.field "name" Json.Decode.string)
+        -- I choosed to rely on the `type` since it seems way easier
+        -- (and more reliable) than checking `curated` and `shared`
+        -- booleans
```
Janiczek, line 56: :+1: Makes sense!

src/Data/Audience.elm
```diff
@@ -36,6 +40,45 @@ type alias Audience =
...
 
 
+-- Decoders
+
+
+decoder : Decoder Audience
+decoder =
+    -- I know about Json.Decode.Pipeline but for small projects, I prefer to
+    -- limit the number of dependencies. Plus, using `mapX` leads to easier
+    -- to understand error messages than the applicative version.
```
Janiczek, line 50: Good to see you know about the applicative pattern even if you chose not to use it here. (By the way, in our main codebase we've made the choice to do all decoders via `Decode.succeed Record |> Decode.andMap ... |> Decode.andMap ...` even if they fit into the `mapN` limit - just so that the codebase is consistent and any irregularities are very visible if the API is weird in some way.)

Can you expand on the difference in error messages between applicative and `mapX`? That's interesting to hear. Perhaps one shows "number of arguments" error and the other one "wanted to get `X`, but got `Y -> X`" or something similar? :eyes:

src/Data/Audience.elm
```diff
@@ -36,6 +40,45 @@ type alias Audience =
...
+                        Json.Decode.succeed Authored
+
+                    "shared" ->
+                        Json.Decode.succeed Shared
+
+                    "curated" ->
+                        Json.Decode.succeed Curated
+
+                    _ ->
+                        Json.Decode.fail ("invalid audience type: " ++ typeStr)
```
Janiczek, line 77: Showing the bad value - nice :+1:

src/Data/AudienceFolder.elm
```diff
@@ -7,10 +10,14 @@ This module implements everything related to audience folder resource.
 
 # Interface
 
-@docs AudienceFolder, audienceFoldersJSON
+@docs AudienceFolder, fixtures
```
Janiczek, line 13: Looks like the doc comment is a bit useless given that we don't `elm make --docs` in our production codebase :thinking:

(I wanted to comment on the fact that `decoder` fn is missing from the doc comment, but I don't think it's a very big issue -- precisely because we don't generate the docs...)

Well, food for thought - we can probably tweak our challenge repo a bit after all the hiring dust settles :)

src/Main.elm
```diff
@@ -1,10 +1,396 @@
...
+    case ( msg, model ) of
+        ( LoadingMsg loadingMsg, Loading loading ) ->
+            updatedLoading loadingMsg loading
+
+        -- instead of using the `_` wildcard for all the other possible
+        -- (msg, model) combinaison, I pattern match against LoadingMsg and
+        -- "anything else". This way, if I add some variant to the Msg
+        -- type, the compiler will help me finding out where to add code.
+        ( LoadingMsg _, _ ) ->
+            ( model, Cmd.none )
```
Janiczek, line 71: Ahh, [where have I seen this](https://github.com/elm-in-elm/compiler/pull/68) :smile: 

Nice!

src/Main.elm
```diff
@@ -1,10 +1,396 @@
...
+{-| simulates a http loading
+-}
+wait : Msg -> Cmd Msg
+wait msg =
+    Task.perform (always msg) (Process.sleep 1000)
+
+
+listInData : Decoder a -> Decoder (List a)
+listInData decoder =
+    Json.Decode.field "data" (Json.Decode.list decoder)
```
Janiczek, line 35: An opinion: I like it more if all these decoder field traversals are inside a single function, all at once, to keep it understandable and not have to chase the magic around various functions and modules -- so I'd keep this inline in the `Http.get` call or wherever the top-level decoder for `List Audience` has to live.

src/Main.elm
```diff
@@ -1,10 +1,396 @@
...
-main : Html msg
+type Model
+    = Loading LoadingModel
+    | DecodingError
+    | Loaded LoadedModel
+
+
+type Msg
+    = LoadingMsg LoadingMsg
+    | LoadedMsg LoadedMsg
```
Janiczek, line 23: Could you please walk me through your reasoning on why you chose this approach (custom type instead of a record; Msgs separated by Model cases they apply to), and perhaps pros and cons that you considered between the alternatives?

src/Data/AudienceTree.elm
```diff
@@ -0,0 +1,361 @@
...
+type alias AudienceTreeInternal =
+    { name : String
+    , children : List AudienceTree
+    , audiences : List String
+    }
+
+
+children : AudienceTree -> List AudienceTree
+children (AudienceTree a) =
+    a.children
```
Janiczek, line 29: Nitpick: The name `children` was a bit confusing to me when I read it in the `String.fromInt <| List.length (Tree.children tree) + List.length (Tree.audiences tree)` snippet in `Main`. Maybe a name like `childTrees` might describe it better - make it clear that no audiences are present in the result value?

src/Main.elm
sebsheep: I use this trick as much as I can now since I discovered it in elm-in-elm!

src/Main.elm
sebsheep: I use a sum type for the model instead of a product one because here.... well... my model is better represented by a sum type! I cannot display anything to user before I have all the data and the user cannot do anything neither. So "Loading" is a special state.

I thinked about it a few minutes, but I only see benefits with this approach and no drawbacks. Because if all the msgs were "flat", we'd have to do something like:
```
update msg model =
    case msg of
        FolderClicked ->
                 -- it doesn't mean anything to open a non loaded folder

                  if isLoaded model) then
                       ....
                  else 
                      <noOp>
        GotFolders (Ok folders) ->
                  if not (isLoaded model) then
                      ....
                  else
                     <noOp>
       ....
```
leading to a lot of repeating code. But if you ask, you surely see some drawbacks and I'd be happy to hear your point of view about that.

src/Data/AudienceTree.elm
sebsheep: Good catch, you're right!

I'm not really happy with the wording for this module. Maybe `subFolders` might be even better than `childTrees`.

src/Data/Audience.elm
sebsheep: Yeah, that is mainly because of the "wanted `X` but got `Y -> X`" which are quite cryptic. 

E.g, if you add a field to a record `MyRecord` and you have some decoder:
```
myRecordDecoder = 
   Decode.succeed MyRecord
      |> ....
```
you'll got this hard to understand error. But if you had:
```
myRecordDecoder =
    Decode.map3 MyRecord
     |> ...
```
the compiler can immediately see that `map3` and `MyRecord` don't match. But yeah, it is a matter of taste :) 

(and in production apps, I tend to use `Json.Decode.Pipeline`)

sebsheep: Related: I've recently discovered that "uniformity" is not such an important goal and can easily lead to "rigidity", which is clearly not desirable. I did a long time deeply understand what Evan meant with this: https://guide.elm-lang.org/webapps/modules.html#growing-modules .

src/Data/AudienceTree.elm
```diff
@@ -0,0 +1,361 @@
...
+-- In this step, we add all the audiences to their corresponding PendingFolder.
+-- If the folder indicated in the audience record doesn't exist, the audience
+-- is assigned to root and we keep track of this audience in the
+-- `missingFolderId` list.
+--
+-- This step has to be performed AFTER the previous one (we have to know all the
+-- folders ids). This is "softly constrained" by the fact that the `step2`
+-- function takes `Step1Result` as argument (we might use some opaque type technique
+-- to have a "hard constaint" about function call order but this would split the process
+-- across multiple files which doesn't seem to worth it for such a small project).
```
Janiczek, line 163: > which doesn't seem worth it for such a small project

Honestly, even your three-step process with the error handling is quite a bit of overkill IMHO :joy: In normal code I'd perhaps consider relying on BE guarantees more (no duplicities, no folders/audiences with parent that doesn't exist) but it's cool that you're showing this off!

src/Main.elm
```diff
@@ -1,10 +1,396 @@
...
+                            [ Html.text (Data.AudienceTree.name loadedModel.currentTree) ]
+                       ]
+           )
+        |> List.intersperse (Html.div [ Html.Attributes.class "mx-1" ] [ Html.text ">" ])
+        |> Html.div [ Html.Attributes.class "flex flex-row text-sm items-center" ]
+
+
+type alias SplittedBreadcrumb =
+    { currentTree : AudienceTree
+    , breadcrumbs : List AudienceTree
```
Janiczek, line 369: Does this hold large parts of the whole tree in memory as you navigate further into the tree?
eg.
```elm
-- full tree = degenerate 1-child tree looking like [A,B,C,D,E,F,G,H]
{ currentTree = [D,E,F,G,H]
, breadcrumbs =
  [ [C,D,E,F,G,H]
  , [B,C,D,E,F,G,H]
  , [A,B,C,D,E,F,G,H]
  ]
}
```
What would you consider as alternatives if there was so much data in the full tree that this duplication started being a problem?

----

EDIT: just to put this in perspective, in our production my user sees: 445 audiences (but some people go bonkers, having thousands), and 6 audience folders (but some people create more, let's say <100).

But there are other entities ("questions" and "categories", which are like folders for questions) where you see high thousands of questions (my user: 8278) and thousands of categories (my user: 1965).

I suspect this duplication might not be a problem in case of audiences+audience folders, but it most likely would in case of questions+categories. So, what to do instead?

----

EDIT 2: going against my initial comment is this thread on Elm Slack: https://elmlang.slack.com/archives/C0CJ3SBBM/p1601593754295100
TL;DR: there is structural sharing happening.

public/index.html
```diff
@@ -0,0 +1,21 @@
...
+    <title>Audience Folders</title>
+    <script src="main.js"></script>
+    <link href="https://unpkg.com/tailwindcss@^1.0/dist/tailwind.min.css" rel="stylesheet">
+    <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">
+</head>
+
+<body>
+    <div id="myapp"></div>
+    <script>
+        var app = Elm.Main.init({
```
janjelinek, line 14: You don't need `var app = ` here.

elm.json
```diff
@@ -3,15 +3,15 @@
     "source-directories": [
         "src"
     ],
-    "elm-version": "0.19.0",
+    "elm-version": "0.19.1",
```
janjelinek, line 6: 💪

src/Data/Audience.elm
janjelinek: It make sense from information we give you in this challenge, but ignoring those booleans means you will discard combinations like
`shared = True` and `curated = True` with `type = Authored`.  
But no big deal in this case 👍

src/Data/AudienceFolder.elm
```diff
@@ -64,6 +83,18 @@ audienceFoldersJSON =
...
+                "id": 3111,
+                "name": "New Group 2 - Duplicated",
+                "curated": false,
+                "parent": 3110
+            },
+            {
+                "id": 12345,
+                "name": "Parent doesn't exist",
+                "curated": false,
+                "parent": 42
```
janjelinek, line 97: I like your thinking about invalid states here, but I also think, that some logic and cases should be handled only in one place and in this case on API side so you will never get such an invalid data. This means less code on FE. But we do not live in ideal world so 👍

src/Data/AudienceTree.elm
```diff
@@ -0,0 +1,361 @@
...
+E.g. if an audience folder references a non-existing folder, this audience is directly
+added to the root, and this audience is added to the `missingFolderId` field.
+-}
+fromAudiencesAndFolders : List Audience -> List AudienceFolder -> ( AudienceTree, Errors )
+fromAudiencesAndFolders audiences_ folders =
+    let
+        step3result =
+            step1 folders
+                |> step2 audiences_
+                |> step3
```
janjelinek, line 60: I don't like those generic names `stepN` and even fact, that those function are defined below their usage, but I don't have better naming for you know and "function definitions below it's first usage" is something I fighting with even in our production code :D

src/Data/Audience.elm
```diff
@@ -5920,6 +5963,17 @@ audiencesJSON =
...
+            {
+                "id": 91958,
+                "name": "My Folder doesn't exist",
+                "expression": {
+                },
+                "curated": null,
+                "type": "shared",
+                "shared": true,
+                "category": null,
+                "folder": 9999999
```
kraklin, line 5976: Nice fixture, we should bake this permanently into the original challenge data 👍

src/Data/AudienceTree.elm
```diff
@@ -0,0 +1,361 @@
...
+-- It would be problematic if multiple folders share the same id `I`. If an audience
+-- belongs to `I`, to what folder should we put it? If we encounter
+-- such a case, only the first folder is kept and we keep track of those problematic folders
+-- in the `duplicatedFolderIds` list.
+
+
+type alias Step1Result =
+    { root : PendingRootFolder
+    , otherFolders : Dict Int PendingFolder
+    , duplicatedFolderIds : List DuplicatedIdFolder
```
janjelinek, line 85: This is super overkill I would say. It's nice, but I thinks (as I mentioned already) that this has to be handled on BE. Especially if you have control over it (API is own by your team or team in your company). 
And also it generates 4 extra lines of comments and I don't like too much comments :)

src/Main.elm
janjelinek: It seems OK to me. I was thinking about future request such a "be able reload already loaded audience" but it's relatively easy achievable with you approach 👍

janjelinek (commented):
I like your thinking about invalid/error states even if it's maybe too much in some cases. It's clearly see that you know what you doing and it's better think about more cases than underestimate something. 

I don't see any CSS here, you did clever trick here with 4rd party styling framework so we don't anything about your CSS skills now. 

Overall Elm part is very good, CSS is non existing, build is even optimised, but you committed build output (not necessary, even not needed).

Good work 👍

src/Data/AudienceTree.elm
kraklin: Having the function defined before it's usage is something I would like to see more in our codebase as well, it is easily searchable this way in Vim for me :)

kraklin:
There is nothing much to be said from my point of view as others have said that before me :) I like the code and yours thinking over the problem even though it is really unnecessarily complex for the given task.

All I can say is - job well done @sebsheep

src/Data/Audience.elm
Janiczek: Yes, given the types in this challenge, it doesn't make sense to think about combinations of `shared+curated+user`.
(But in our codebase this has indeed evolved into three separate fields. :sweat_smile: )

src/Data/AudienceTree.elm
Janiczek: Not this flamewar again :D

Janiczek:
> I don't see any CSS here, you did clever trick here with 4rd party styling framework so we don't anything about your CSS skills now.

@janjelinek It's not like it's Bootstrap which would be a separate abstraction on top of CSS.
Tailwind seems to me like a shorthand version of writing normal CSS - all the classes map to single properties. So I **would** extrapolate CSS knowledge from what is shown in the `view` functions.

Janiczek:
@sebsheep Overall I liked that you've shown the ability to do non-trivial recursion-style functions (to construct the tree); but my preference for production code would be to ask product managers about what should we do about the edge cases / ask BE about API guarantees (folders missing; duplicate items or duplicate folders).

I'm _**still**_ interested in hearing your thoughts on selecting this solution between other solutions like tree zipper or a "relational" approach (just keep the two original dicts in memory, current folder ID in model, and do any filtering of the current level in `view`).

TL;DR: I wouldn't choose such a "big" solution for constructing the tree (given the rule `You can use any 3rd-party library you want`, I'd perhaps offload that algorithmic burden to something like [`elm-rosetree`](https://package.elm-lang.org/packages/zwilias/elm-rosetree/latest/Tree#unfold) or [`lazy-tree-with-zipper`](https://package.elm-lang.org/packages/turboMaCk/lazy-tree-with-zipper/latest/Lazy-Tree#build)). But interview code questions are not the same thing as having to implement and maintain the thing for real so don't take this as a negative "rating" :smile: Pretty solid showing! :+1:

src/Main.elm
sebsheep: Yeah, immutable structures are very cool concerning data sharing!!

However, it might be an issue if the folder structureis "linear" with a very long chain of folders "folder1 > folder2 > folder 3 >... > folder1000" since the complexity of building these breadcrumbs is quadratic in the depth of the tree. 

A possible solution would be to limit the breadcrumbs length to say 10 (which already seems a huge depth for a folder structure so I'm not sure I'd do that).

public/index.html
sebsheep: right, but in real app, it is very likely we'd use ports so :man_shrugging:

src/Data/AudienceTree.elm
sebsheep: I don't like those names neither... I searched for better names but didn't succeed and finally decided it was private things and it wasn't a big deal (yeah I'm lazy sometime...).

sebsheep: And about "function definitions above it's first usage", I'd say it is an historical technical constraint. My thinking about this is "first describe the problem at high level, then give implementation details". 

That being said, with the "Go To Definition" feature accessible with one key press, I think declaration order doesn't really matter since with large files, we rarely read them linearly.

sebsheep:
Hi folks!

Thanks for all your feedback, I rarely had so much comments! Below are some answers to previous comments/questions. Feel free to ask me things again if I missed something.

**Error handling, BE guarantees**

First of all, I have to that confess this part is kind of "look what I'm able to do" and I'm not sure I'd write this precise error handling in production.

 However, if the BE can guarantee all the things I check in the code (roughly, we have a tree structure without "orphans"), why wouldn't it serve those data as a tree? It would make it way easier to "paginate" in case of enormous folder structure. <troll>Plus, if we blindly trust server guarantees, why would we write decoders?</troll> More seriously, I'd actively think about make the server responding such a tree (or speak with the BE team if I don't have hand on BE).

Moreover, I'd be really happy to know if you can guarantee tree structure validity with SQL constraints. If we cannot do that, we need *something* to check this structure, and this *something* would be close to what I've written. So I think this code has to be "somewhere" (whether BE-side or FE-side this is implementation detail).

Finally, if folders and audiences come from distinct services (which seems kind of weird for this precise use case but might have sense in other situations), the FE is *the one* responsible for joining, thus the FE has to check data integrity and perform those checks.

(and of course in "real world" I'd ask to product manager what to do with those duplicate/missing items/folders).

**Why tree over "relational"?**

1. Mainly because in my humble opinion, the `Model` type has to reflect the actual "business" as close as possible. If we do that:
      * the view is absolutely trivial to write
      * we can easily compute additional data if we need it (e.g. like the number of elements in a folder or maximal depth of the tree)
2. Secondary for performance concerns*. If we have a big amount of files in a bunch of multiple folders AND if we want to frequently redraw (e.g. to have some kind of complicated animation asking for elm-side computation) filtering those files at each frame could make the app quite slow.

About "tree zipper" I think I kind of implemented  it with my `breadcrumbs`. I think it is not an actual one because we cannot move to "next sibling"  (but I don't really care since I don't need it).

*yeah, yeah, I konw "premature optimization.... evil... end of the world ..." but it is "Secondary", and it doesn't conflict my first point so i'd say it is a good point.

**CSS**
As pointed by Martin, Tailwind properties merely reflect CSS ones. That said, I'd have to say I don't have strong skills about "build consistent set of CSS classes", but I think I don't think I need it to be really efficient. If I want to share a complex styling property between multiple graphical components, I just create a `Html.Attribute` variable. E.g, if I want a "card" with a red background, some padding (`p-4`) and rounded corner:
```
cardClass = class "bg-red-500 p-4 w-64 rounded-md"
```
and I only use elm to "share" the style (with all the benefits we all know about using elm). 

In my current project I have 40 lines of CSS and the taildwing.config.js file is 47 lines long (total count is about 25K elm loc).

**Using 3rd party lib to build the tree**

Using such libs come at the cost of flexibility. For example, we couldn't change the `children` name to `subFolders` as we discussed if we used one of this solution (or we'd need to write some "wrapper" which defeats a bit the "use it out of the box" purpose). Error handling during building the tree also ask some kind of specific knowledge about what the implicit guarantees are (and what errors you want to log).

All the "trees" have their own peculiarities and I think those "peculiarities" would quickly overwhelm the common-well-known-and-documented algorithms about trees (I didn't know the algorithm to efficiently transform an "relation list" into a tree using immutable structures, it took me about 15 minutes to find and understand it). I think this is specially true if we are thinking "long term".


**Final note**
All of that are my personal convictions and justifications about "why" I wrote the code like this. Of course I'm open to discuss and be very happy to learn I'm wrong. Moreover in "production code", I know to mitigate those convictions to be able to work in a team.

src/Data/AudienceTree.elm
janjelinek: It's not entirely true and this approach is forcing other to use advanced IDE tools as you are using. For you order doesn't matter so is that big issue follow rule "declaration first" which could help others? And also think about code review in GH interface where is usual to follow code from top to bottom by just scrolling and jumping from top to bottom and back during code review is something annoying, trust me :) 
But yeah, stop this flame, this is not the place for it, peace 🦄 and 🌈. Also 🥦 for all.

janjelinek:
> > I don't see any CSS here, you did clever trick here with 4rd party styling framework so we don't anything about your CSS skills now.
> 
> @janjelinek It's not like it's Bootstrap which would be a separate abstraction on top of CSS.
> Tailwind seems to me like a shorthand version of writing normal CSS - all the classes map to single properties. So I **would** extrapolate CSS knowledge from what is shown in the `view` functions.

Yeah, it's like jQuery for JS. Who knows only jQuery is not JavaScript programmer for me :) But I don't think it's that crucial and I believe knowing Tailwind proves some level of CSS world knowledge 👍

PR #18, opened 2020-10-04
Implemented FIRST_STEP and SECOND_STEP.

lawsdontapplytopigs:
Thank you for the amazing challenge. It was a really fun exercise!
This is my implementation of both "FIRST_STEP.md" and "SECOND_STEP.md".

# Building
I've also included a build script. For a dev build just run `./build.sh`, which will create a "built/" directory. You can run `elm reactor` in that directory and go to "localhost:8000/index.html" in your browser to see the app.

# Notes
As opposed to what "FIRST_STEP.md" says, after discussing with @Janiczek on slack, I decided to use the mdgriffith/elm-ui library for rendering the ui. I have left a comment as to why I've made that decision in the "src/Main.elm" file.

# Potential improvements
* For an optimized build, we could also minify the "built/app.js" file.
* Use elm/svg for the loading animation and perhaps an animation library like mdgriffith/elm-animator if we decide to add many more animations in the future

Should there be something unclear, please let me know either here through comments, or on slack @lawsdontapplytopigs!


src/Data/Audience.elm
```diff
@@ -34,6 +43,63 @@ type alias Audience =
...
+audienceTypeToString t =
+    case t of
+        Shared ->
+            "Shared"
+        Curated ->
+            "Curated"
+        Authored ->
+            "Authored"
+
+audienceTypeToIcon : AudienceType -> Html.Html msg
```
janjelinek, line 56: This is view logic so I would consider to do not have it inside Data module.

src/Data/Audience.elm
```diff
@@ -34,6 +43,63 @@ type alias Audience =
...
+    let
+        typeToAudienceType t =
+            if t == "curated" then
+                JD.succeed Curated
+            else if t == "shared" then
+                JD.succeed Shared
+            else if t == "user" then
+                JD.succeed Authored
+            else
+                JD.fail <| """valid `type` strings are either "curated", "shared", or "user". Instead, we got: """ ++ t
```
janjelinek: Why if/else here? With this amount of branches it's `case` probably better.

src/Data/Audience.elm
```diff
@@ -34,6 +43,63 @@ type alias Audience =
...
+    -- in this case `pure` is Json.Decode.succeed, and `apply` is elm's pipe operator (|>),
+    -- as defined in the NoRedInk/elm-json-decode-pipeline package (which we're using here)
+
+    -- this way we can decode an arbitrary number of additional fields, if we need to
+    (JD.succeed Audience)
+        |> JDP.required "id" JD.int
+        |> JDP.required "name" JD.string
+            -- the "type" field seems most reliable, so I used that one
+        |> JDP.custom audienceTypeDecoder
+        |> JDP.required "folder" (JD.nullable JD.int)
```
janjelinek, line 103: ❤️

Janiczek:
One thing that stands out to me: I miss elm-format :)

src/Main.elm
```diff
@@ -1,10 +1,829 @@
...
+import Process
+import Task
+import Time
 
 
-main : Html msg
+
+-- we could perhaps model this so that you can only make a "FolderClicked" Msg
+-- or "SortBy" Msg only after the model has loaded up, but I don't think this is
+-- too much of an issue
```
Janiczek, line 53: Our codebase takes similar approach. We rarely have a custom-type Model with a separate Msg for each Model case. Instead we have Model as a record and all Msgs together -- this means sometimes we have to do `if foo model then (model, Cmd.none)` if the Msg is only relevant in some cases, but the developer ergonomics are IMHO better than with custom-type Model and "sub-Msgs"

src/Main.elm
```diff
@@ -1,10 +1,829 @@
...
+    = FolderClicked Int
+    | GoBack
+    | SortBy Data.Audience.AudienceType
+    | GotAudiencesJSON String
+    | GotAudienceFoldersJSON String
+    | GotTime Time.Posix
+
+
+
+-- we use dicts instead of lists because it gives us O(log n) time complexity
```
Janiczek, line 66: I agree with the whole comment :+1:

src/Main.elm
```diff
@@ -1,10 +1,829 @@
...
+    { audiences : Dict.Dict Int FormattedAudience
+    , audienceFolders : Dict.Dict Int FormattedAudienceFolder
+
+    -- we store the full path, because we want to record the full history
+    -- to go back in the tree of folders
+    , currentFolderPath : List Int
+
+    -- according to "SECOND_STEP.md" there must be at least one category selected
+    -- at any time
+    , sortBy : Data.Audience.AudienceType
```
janjelinek: In fact it's not sorting, but filtering.

src/Main.elm
```diff
@@ -1,10 +1,829 @@
...
+-- we fake an http request with a delay
+
+
+fakeHttpGet : Msg -> Float -> Cmd Msg
+fakeHttpGet msg time =
+    Task.perform (\_ -> msg) (Process.sleep time)
+
+
+init : () -> ( Model, Cmd Msg )
+init flags =
```
janjelinek: Unused `flags`, could be `_`

src/Main.elm
```diff
@@ -1,10 +1,829 @@
...
+    }
+
+
+tryLoadingModelToLoaded : LoadingModel -> Model
+tryLoadingModelToLoaded loadingM =
+    case ( loadingM.audiences, loadingM.audienceFolders ) of
+        ( Just loadedAudiences, Just loadedAudienceFolders ) ->
+            Loaded (initLoadedModel loadedAudiences loadedAudienceFolders)
+
+        ( _, _ ) ->
```
janjelinek: `_ ->` is enough here.

src/Main.elm
```diff
@@ -1,10 +1,829 @@
...
+
+type Model
+    = Loading LoadingModel
+    | Loaded LoadedModel
+    | JsonError Json.Decode.Error
+
+
+type alias LoadingModel =
+    { audiences : Maybe (Dict.Dict Int FormattedAudience)
+    , audienceFolders : Maybe (Dict.Dict Int FormattedAudienceFolder)
```
Janiczek: In my mind, using WebData from the RemoteData package might remove the need for having one model type with Maybe Dict and another with Dict.

FYI in our codebase we have mostly the WebData approach, but there are some places where we go with what you did - two types, one with Maybe and one "we surely have the data now".

src/Main.elm
Janiczek: Nitpick: could be `_` just as well; I don't like the wildcard `_` pattern very much in general as it turns off useful compiler errors later down the road when you expand your types due to some product enhancements. (Not that the Maybe type here is going to be expanded.) I'd probably go with `Maybe.map2`  instead here.

src/Main.elm
```diff
@@ -1,10 +1,829 @@
...
+                    , currentFolderPath = []
+                }
+
+        ( GotTime time, Loading loadingModel ) ->
+            Loading
+                { loadingModel
+                    | time = Just time
+                }
+
+        ( _, _ ) ->
```
janjelinek, line 240: This is always danger for future maintenance, you can easily forget cover some new combination.

src/Main.elm
```diff
@@ -1,10 +1,829 @@
...
+            Loaded (initLoadedModel loadedAudiences loadedAudienceFolders)
+
+        ( _, _ ) ->
+            Loading loadingM
+
+
+
+-- normally, here I would return a tuple of type (Model, Cmd Msg), but since
+-- in this app there's no Cmd s to run, I decided to simplify everything
+-- and just return the Model
```
Janiczek, line 160: :+1:

src/Main.elm
```diff
@@ -1,10 +1,829 @@
...
+
+
+update : Msg -> Model -> Model
+update msg model =
+    case ( msg, model ) of
+        ( GotAudiencesJSON json, Loading loadingModel ) ->
+            let
+                audiencesDecoder : Json.Decode.Decoder (List Data.Audience.Audience)
+                audiencesDecoder =
```
Janiczek, line 169: I'd move this to Data.Audience!

src/Main.elm
```diff
@@ -1,10 +1,829 @@
...
+                   , E.centerY
+                   ]
+                   <| E.text "hello"
+           ```
+           Which will center the text.
+           With css, even a simple task like this is so unintuitive that there's a
+           very popular thread on stack overflow on how to do it, with answers of
+           varying complexity that I would argue are way too cryptic for the
+           simplicity of the problem.
+           [The thread](https://stackoverflow.com/questions/114543/how-to-horizontally-center-a-div)
```
janjelinek, line 307: Hallo `elm-ui` salesman :)

src/Main.elm
```diff
@@ -1,10 +1,829 @@
...
+
+
+-- normally, here I would return a tuple of type (Model, Cmd Msg), but since
+-- in this app there's no Cmd s to run, I decided to simplify everything
+-- and just return the Model
+
+
+update : Msg -> Model -> Model
+update msg model =
+    case ( msg, model ) of
```
Janiczek, line 165: I'm not used to this (and again the wildcard pattern at the bottom of `update` makes my spidey-sense go tingling :rotating_light:) but given you chose to model the Model as a custom type this is just a consequence of that.

If going this way, I would strongly suggest to follow each `(FooMsg, TheRightModel) ->` with a `(FooMsg, _) ->` so that you at least get *some* compiler checking back (ie. checks that you didn't forget a Msg).

src/Main.elm
```diff
@@ -1,10 +1,829 @@
...
+    }
+
+
+type alias LoadedModel =
+    { audiences : Dict.Dict Int FormattedAudience
+    , audienceFolders : Dict.Dict Int FormattedAudienceFolder
+
+    -- we store the full path, because we want to record the full history
+    -- to go back in the tree of folders
+    , currentFolderPath : List Int
```
Janiczek: I would appreciate a note about the direction of this list in a comment nearby. Is it `[root, child, grandchild]` or `[grandchild, child, root]`?

src/Main.elm
janjelinek: Btw have you ever tried to find some element in code when you inspected it first via browser dev tools? It's really hard with elm-ui. I'm not agains elm-ui, just saying.

src/Main.elm
```diff
@@ -1,10 +1,829 @@
...
+           Also, even if we used elm/html, and we had css separately, if javascript
+           doesn't run, no html nodes will even be created, so again, nothing will
+           be displayed. So both approaches suffer from the same problem.
+
+   It's difficult to convey through these arguments how much more joyful I am
+   styling my app in elm-ui versus css, but it's such a big deal for me that this
+   library was one of the main reasons I picked up elm.
+
+   Also relevant, Matt Griffith's talk about the elm-ui library:
+       https://www.youtube.com/watch?v=Ie-gqwSHQr0
```
Janiczek, line 326: Thanks for the huge rationale comment! <3

src/Data/Audience.elm
Janiczek: To me this depends more on the application. It *might* make sense to put view stuff in Data modules. In our production code we also have some `Data.Foo.toString` or `Data.Foo.label` that are for users to see, not for some decoder.

So I'm undecided about this one. It doesn't scream to me "I belong in Main!" as much :D

src/Main.elm
```diff
@@ -1,10 +1,829 @@
...
+                , E.width E.fill
+                , E.spacing 15
+                , EFont.color Palette.gray0
+                , E.mouseOver <| [ EBackground.color <| E.rgb255 255 255 255 ]
+                ]
+                [ E.el
+                    [ E.width <| E.px 17
+                    , E.height <| E.px 17
+                    ]
+                    <| E.html <| Data.Audience.audienceTypeToIcon audience.type_
```
Janiczek, line 621: Just a curious question as you have more experience with elm-ui than me. How often do you find you need the E.html or E.htmlAttribute escape hatch? In my toy project I found I need it quite a lot for some layout tasks around eg. textareas etc... - elm-ui didn't allow me to flex-grow it as I wanted etc... Is this common?

src/Main.elm
```diff
@@ -1,10 +1,829 @@
...
+            , EFont.bold
+            , EFont.color Palette.gray0
+            ]
+            <| E.text "Loading..."
+        ]
+
+
+
+-- food for thought: Correct me if I'm wrong, but I think E.Element, under this
+-- "nest" operation is a monoid
```
Janiczek, line 705: Makes sense! I wonder about breaking some purity laws with CSS side effects though.

src/Main.elm
```diff
@@ -1,10 +1,829 @@
...
+   to check for all of these potential errors would bring a lot of complexity
+   code-wise and also time-complexity wise. So instead I concluded that this may be
+   better addressed on the backend by writing very good tests.
+   This way our frontend is faster, easier to understand and our tests for these issues
+   only live in one place: the backend.
+
+   Additionally, this made me think that since there are so many things that could go
+   wrong by having this model of references between the audiences and audienceFolders,
+   maybe this would even better be addressed not by writing really good tests on the
+   backend, but instead by data modelling and representing these relationships in a
```
Janiczek, line 776: What would be a good type representation (instead of the relational "two separate tables with `audience.folder_id -> audience_folder.id` link"?)

src/Main.elm
```diff
@@ -1,10 +1,829 @@
...
+   better way, thus making these invalid states impossible to even represent, so we
+   wouldn't even need to test for them.
+
+   Relevant: Richard Feldman's talk "Making Impossible States Impossible"
+   https://www.youtube.com/watch?v=IcgmSRJHu_8
+
+   Just for the record: In production code I wouldn't write these sorts of very long
+   comments. I'd instead probably open up an issue or ping the team on slack.
```
Janiczek, line 784: :smile: :+1:

src/Main.elm
```diff
@@ -1,10 +1,829 @@
...
+-}
+
+-- I haven't spent much time styling this because this is something the user shouldn't
+-- even see. No point in styling it just for it to look cute for the developers
```
Janiczek, line 788: If it's possible for it to happen, it will happen. Backend cache inconsistencies, audiences containing folder IDs for folders that don't exist... we've seen it all :smile:

src/Main.elm
```diff
@@ -1,10 +1,829 @@
...
+        <| E.column
+            [ E.width E.fill
+            , E.height E.fill
+            ]
+            [ E.el [ E.paddingXY 0 80, E.centerX, E.height E.fill ] content
+            , E.row
+                [ E.width E.fill
+                , E.alignBottom
+                ]
+                <| List.map viewColor colors
```
janjelinek, line 364: So if I'm right, this piece, with some code in Palette and with `colors` list is producing bunch of elements aligned by `flex` to produce basically this CSS baground, right?
```CSSS
background: linear-gradient(90deg, rgba(10,30,60,1) 0%, rgba(10,30,60,1) 16.6%, rgba(20,46,80,1) 16.6%, rgba(38,74,120,1) 33.2%, rgba(20,46,80,1) 33.2%, rgba(38,74,120,1) 49.8%, rgba(90,150,225,1) 49.8%, rgba(90,150,225,1) 66.4%, rgba(130,180,235,1) 66.4%, rgba(130,180,235,1) 83%, rgba(192,228,255,1) 83%, rgba(192,228,255,1) 100%);
```
Not sure which approach is more obvious from code 🤷

Janiczek (commented):
Pretty solid! I like the overall approach to breadcrumbs and navigation. I've left some comments and questions inline. Thanks for helpful comments that help us see how you think and what alternatives you considered!

src/Main.elm
Janiczek: @janjelinek for the record when @lawsdontapplytopigs asked me about whether he can use elm-ui instead of CSS, I told him to give us some rationale / explanatory comment. So this is it. I asked for it :)

Janiczek: Also, the question about mapping actual DOM back to code is a good one and I'm curious about answers to that. I only know about Element.explain which does something similar but in the other direction (you add it to some view code and it gives a distinctive border to the element and its children IIRC).

You could iteratively drill down to the wanted DOM element that way but it seems like a worse solution than having BEM-like classes around and navigating by these.

src/Main.elm
```diff
@@ -1,10 +1,829 @@
...
+viewSortingPanel : Data.Audience.AudienceType -> E.Element Msg
+viewSortingPanel currentlySortBy =
+    let
+        buttons =
+            List.map
+                (\btn -> btn currentlySortBy)
+                [ sortButton Data.Audience.Authored
+                , sortButton Data.Audience.Curated
+                , sortButton Data.Audience.Shared
+                ]
```
janjelinek:
```suggestion
            List.map
                (\btn -> sortButton btn currentlySortBy)
                [ Data.Audience.Authored
                , Data.Audience.Curated
                , Data.Audience.Shared
                ]
```
Or you can go with point free if you swap arguments for `sortButton` function.

src/Main.elm
```diff
@@ -1,10 +1,829 @@
...
+                    []
+
+                _ ->
+                    model.audienceFolders
+                        |> Dict.filter (\key value -> value.parent == currentFolder)
+                        |> Dict.foldl viewFolder []
+
+        filteredAudiences : List (E.Element Msg)
+        filteredAudiences =
+            case model.sortBy of
```
janjelinek: See? You knew it's filtered, but still you have it named `sortBy` :)

src/Main.elm
```diff
@@ -1,10 +1,829 @@
...
+                Data.Audience.Shared ->
+                    model.audiences
+                        |> Dict.filter (\key value -> value.type_ == model.sortBy)
+                        |> Dict.foldl viewAudience []
+
+                _ ->
+                    model.audiences
+                        |> Dict.filter (\key value -> value.folder == currentFolder)
+                        |> Dict.filter (\key value -> value.type_ == model.sortBy)
```
janjelinek: This can be done in one iteration

src/Main.elm
```diff
@@ -1,10 +1,829 @@
...
+                ]
+                [ E.paragraph
+                    [ E.paddingXY 5 0
+                    ]
+                    [ E.text "Currently sorting by "
+                    , E.el
+                        [ EFont.bold ]
+                        <| E.text (Data.Audience.audienceTypeToString model.sortBy)
+                    ]
+                , wrapColorCircle Palette.color3 coolI -- we wrap it in a colored circle
```
janjelinek: Sorry and no offence, I just had to :D 
thank-you-captain-obvious

src/Main.elm
```diff
@@ -1,10 +1,829 @@
...
+        , viewSortingPanel model.sortBy
+        ]
+
+
+viewFolder : Int -> FormattedAudienceFolder -> List (E.Element Msg) -> List (E.Element Msg)
+viewFolder id folder acc =
+    let
+        -- we view a "folder" as a row: the icon, and the name of the folder
+        item =
+            E.row
```
janjelinek, line 583: It's nice, that elm-ui is generating `role` for `button` types, but in case of list of items it's not that accessibility friendly for users as it looks from Elm code base.

src/Main.elm
```diff
@@ -1,10 +1,829 @@
...
+            E.el
+                [ E.centerX
+                , E.centerY
+                ]
+                <| E.text formattedErrText
+
+
+
+-- intermediate representation for "audience"s and "audienceFolder"s , so we
+-- can easily make dicts out of them
```
janjelinek, line 813: What's wrong about usage Audience directly and have it converted to dict this way
```Elm
audiences
        |> List.map (\a -> ( a.id, a ))
        |> Dict.fromList
```
It's more beneficial for future changes in data structure. If Audience will be extended let's say by `createdAt : Posix` you need to change this on two places instead of one. It's not big chunk of work, but what if you will have more modules where `Data/Audiences` are used. It will require at least share `FormattedAudience` too.

src/Main.elm
janjelinek, line 784: 👍 for your thoughts here. Anyway what about empty folders? Do you think about them? In your implementation are empty folders always shown even if they are clearly belongs to `Curated` audiences only.
Also about complexity and covering impossible states, did you consider use some precomputed data structure, where you will be constructing whole `tree` only once (when data are loaded) instead of checking if audience belongs to currently selected folders in every view cycle?

janjelinek (commented):
First, thx for you PR and effort!

I see lot of thinking here and good looking implementation. 
I'm still not entirely sold for elm-ui usage in bigger projects, but for this scale it's pretty good. 

I commented mostly only part's which from my point of view could be done differently/better but in general 👍

src/Data/Audience.elm
lawsdontapplytopigs: Yeah I thought it wasn't that big of a deal and my thought process was "ok so this is a function that works on `AudienceType`, so it makes sense to put that function there." (I thought this was a good idea after watching Evan's talk [the life of a file](https://www.youtube.com/watch?v=XpDsk374LDE))

src/Data/Audience.elm
lawsdontapplytopigs: I usually use if/else with strings because for some reason it got stuck in my head from the very beginning that "oh no, these are strings, so you're going to have to add `case` expressions for every possible string ever." Even though you can obviously just add a "_" at the bottom to cover any other case. But I will consider changing my ways. Having a case expression is probably a bit less noisy.

src/Main.elm
lawsdontapplytopigs: You're right, my mistake.

src/Main.elm
lawsdontapplytopigs: What I actually do is not even that, but instead, wherever I'm unsure of where an element is positioned, how much space it's taking, etc, I add a `EBackground.color <| E.rgb255 80 80 80` property there, re-compile, and see in the browser what that element actually is and how much space it's taking. 
It barely takes time, and then you can easily comment that out and keep it around to un-comment it in the future to see if anything is different.

Usually I don't even go by using the inspector because it takes a bit of fiddling around to find the element you're looking for, whereas just changing an element's background makes it pop out right away. And I also usually don't use my mouse since I use vim. For me this is a win-win, but I'm not so sure how much of the team uses IDEs vs more oldschool editors.

Yeah I know that comment is kind of a pitch, it's just that I really like that library ;;)

src/Main.elm
lawsdontapplytopigs: Yeah, same here. I usually tend to stray away from really nested Msgs because then I usually find it's harder to trace back where something came from. Whereas if it's just a more flat `Msg` type, I find it easier to find stuff in one single file.

src/Main.elm
lawsdontapplytopigs: The reason I did it like this is because I tried to come up with really good looking code, but I agree with you. I don't think this is a good pattern to keep using, and even in my previous projects I have actually always handled every possible `Msg` case individually.

It's just that in this case it got really nested with how I used a union type for the `Model`, so I thought it *looked* better like this.

src/Main.elm
lawsdontapplytopigs: You're right, that would be a bit clearer!

src/Main.elm
lawsdontapplytopigs: hahaaaa now that you mention it, it's super obvious. )))))

src/Main.elm
lawsdontapplytopigs: no problem!! ;;;););)

src/Data/Audience.elm
Janiczek: I believe it all compiles down to the same thing; I guess we prefer `case string of ...` because of how it looks "tabular"?
Anyways I'd consider it idiomatic, but it's not a big deal at all.

src/Main.elm
Janiczek: Just some context: Our need to go from the browser DOM to the code is driven by "Hey we need to change the color of this label!" requests by product managers, after 3 months of not looking at the code. So for us it's faster to open the inspector, look at the class `.xb--browser--search-results--item--icon` and then guess an Elm module by that and search for `"icon"` in it. That gets you to the correct Elm function quite fast. I'm really unsure how we'd do that without these classes.

src/Main.elm
lawsdontapplytopigs: Yeah I almost never use `Element.html`, except for APIs that return `Html.Html msg`.

## Cases where I've used `Element.html`
- `Element.html <| Svg.svg ...`
- `Element.html <| WebGL.toHtml ...`

As for `Element.htmlAttribute`, yeah I've actually used that a bit more.

## Notable cases where I've used `Element.htmlAttribute`
- `Element.htmlAttribute <| Html.Attributes.transform ...`
- `Element.htmlAttribute <| Html.Attributes.style "z-index" ...`
- `Element.htmlAttribute <| Html.Attributes.id ...`
- `Element.htmlAttribute <| Html.Attributes.style "overflow" "hidden"`

Other than that I've actually done everything in elm-ui.

And especially when it comes to laying out text, there are certain "recipes" or guidelines that you use in elm-ui.
For example, to center text:
```elm

view model = Element.layout [] <|
    Element.el
        -- make this element take up the whole page width and height
        [ Element.width Element.fill
        , Element.height Element.fill
        ]
        <| Element.el 
            -- since we don't specify here how much width or height this will take,
            -- it will take up only as much size as the text takes up, so it will
            -- be placed exactly in the middle.
            [ Element.centerX
            , Element.centerY
            ]
            <| Element.text "This text will be centered."
```
In short, I've almost never really needed to use `Element.html` or `Element.htmlAttribute` to do layout, even when it comes to text.

src/Main.elm
lawsdontapplytopigs: oh you're right!! I missed that one

src/Main.elm
lawsdontapplytopigs: It does seem that for a purpose like this, you could use [Element.Region](https://package.elm-lang.org/packages/mdgriffith/elm-ui/latest/Element-Region#heading) to "mark" some nodes, so elm-ui will make them "li", or whatever you specify, although I'll be honest and say I haven't used it yet.

src/Main.elm
lawsdontapplytopigs: Yeah, you're right. My thought was that it was nicer because since we're working with dicts, it'd be a bit clearer to have a function name there that literally says "audiencesToDict" or something like that, but now that you mention it, I agree that things could easily go out of sync, and it'd just make you unnecessarily change code in 2 places, just for the `id` field which is different.

src/Data/Audience.elm
janjelinek: It doesn't hurt me much as well, just opinion, but I would just not expect function with output like `Html msg` in data module 🤷

src/Data/Audience.elm
janjelinek: I was just curious about your preferences and I agree, `case` is less noisy and by that more obvious for the first quick look 👍

src/Main.elm
lawsdontapplytopigs: I have some code I want to show regarding this, but I wanna make sure it's right, so I'll answer this tomorrow because it's late right now

src/Palette.elm
```diff
@@ -0,0 +1,28 @@
+module Palette exposing (..)
```
kraklin, line 1: I'd think of different name if the module contains something else than just colours. It wouldn't be my pick when I'd look for font sizes, spacings etc..

src/Main.elm
```diff
@@ -1,10 +1,829 @@
...
+        -- "load" these at different times, to show that the UI works right
+        -- the user can't do anything with the audienceFolders and audiences
+        -- until both of them load
+        cmd_ =
+            Cmd.batch
+                [ fakeHttpGet (GotAudiencesJSON Data.Audience.audiencesJSON) 300
+                , fakeHttpGet (GotAudienceFoldersJSON Data.AudienceFolder.audienceFoldersJSON) 2000
+                ]
+    in
+    ( model_, cmd_ )
```
kraklin, line 133: There is no need for the underscored suffix in here, but I like it as it can be easily forgotten that you are not working with the "real" models and cmds. Or has there been another thinking behind the `_` suffix?

src/Main.elm
kraklin: I love elm-ui as well and use it on my projects as much as I can, but the problems are when you are not the one doing the design. Once designers steps in, it can be hard to persuade them that what they ask for can't be done with elm-ui.

I have found out that using some kind of BEM like CSS approach can be quite nice middle ground between pure CSS madness and not being able to do things you are asked about. We are using [WeakCSS](https://github.com/GlobalWebIndex/weak-css) and it has been working pretty well for us so far.

src/Main.elm
kraklin: I would prefer the CSS way here as well. Or having this in separate function with some name like "fancy gradient background" would work as well here for me :)

src/Main.elm
```diff
@@ -1,10 +1,829 @@
...
+
+colors : List E.Color
+colors =
+    [ Palette.color0
+    , Palette.color1
+    , Palette.color2
+    , Palette.color3
+    , Palette.color4
+    , Palette.color5
+    ]
```
kraklin, line 659: I'd expect this to be in Palette module as well...

src/Main.elm
```diff
@@ -1,10 +1,829 @@
...
+
+
+viewLoading : LoadingModel -> E.Element Msg
+viewLoading loadingModel =
+    let
+        currentTime =
+            case loadingModel.time of
+                Just t ->
+                    toFloat (Time.posixToMillis t) / 70
```
kraklin: Why it is divided by `70`? (I'm really suspicious towards magic numbers :))

kraklin (commented):
Thank you for your submission Adrian.

I really liked the thinking about the unnecessary complexity of the app in the first place + I like the look of the app and how you have achieved it (elm-ui FTW)

I have just few nitpicks as others have similar questions I would have.

Good job

src/Main.elm
lawsdontapplytopigs: Update: I actually got really curious how accomplishing something like this efficiently would look in code, so I started working on an implementation. I'll probably commit it tomorrow. I hope that's not too late!

src/Main.elm
lawsdontapplytopigs: That number actually controls the speed of the loading animation. But you're right, leaving something like that undocumented can easily cause confusion.

src/Main.elm
lawsdontapplytopigs: The reason I didn't is because this is only used in the Main.elm file. If we had other files, and they wanted to have a list of colors, I think those lists should be put in the module that uses them.

My thought process was "what if we had a module that instead of needing a list of all colors from the Palette module, it only needed `[Palette.color1, Palette.color3, Palette.color5]`? Or it needed them in a different order? Should that belong in Palette as well?"

This is why I concluded that a list of colors belongs in the module that actually uses it.

src/Main.elm
lawsdontapplytopigs: Oh yeah, about that
What I often do is the following:

When I first started, something in the update function would look like this:
```elm
update msg model =
    case msg of
        Msg1 value ->
            ( { model
                | field = value
                }
            , Cmd.none
            )
```
But I really disliked the way that tuple was indented. And also, the last brace in that record update looks off. What looks better? What I wrote above?
Or this?
```elm
update msg model =
    case msg of
        Msg1 value ->
            ( { model
                | field = value
            }
            , Cmd.none
            )
```
or maybe this:
```elm
update msg model =
    case msg of
        Msg1 value ->
            ( 
                { model
                    | field = value
                }
            , Cmd.none
            )
```

So at some point I got tired arguing with myself which looks *slightly* better, so I just did the following everywhere:

```elm
update msg model =
    case msg of
        Msg1 value ->
            let
                model_ =
                    { model
                        | field = value
                    }
                cmd_ = Cmd.none
            in
            ( model_, cmd_ )
```

This is how I came to use the `_` suffix for those two things.

src/Palette.elm
lawsdontapplytopigs: The reason for this module name is actually something I got from Matt Griffith's talk on elm-ui [at 23:35](https://youtu.be/Ie-gqwSHQr0?t=1414), where he basically mentions that it's a good idea to design starting with a palette, which includes colors, along with fonts used, sizing, etc.

I understand this can be easily mistaken to only stand for Palette of colors, so perhaps a more descriptive name would be `UIPalette`.

src/Main.elm
lawsdontapplytopigs: I got curious of this myself, so I started working on an implementation for something like this! I'll most likely be done tomorrow, and I'll submit another commit!

lawsdontapplytopigs:
## Addressing @janjelinek and @Janiczek concern
### "What would be a good type representation (instead of the relational "two separate tables with audience.folder_id -> audience_folder.id link"?)"

I think some sort of tree structure that we hold in memory and can navigate would be the best approach here.
I've left a sample JSON string inside the `src/Data/AudienceTree.elm` file that probably explains this the best, but basically I think we need a tree of some sort.

That way we would solve the following issues:
1. There's no more potential deadlocks in cases of "folder1 points to folder2 and folder2 points to folder1". If these are nested, we can't have that issue anymore. if folder1 contains folder2, folder2 **can't** contain folder1.
2. If the data that comes back from the server is good, an `Audience` can't contain an `AudienceFolder`. In the case where everything works with `id`s, we can have an AudienceFolder whose `parent` field can point to an `Audience` (if they're in the same Dict)
3. What if an audience's `folder` field points to a folder that doesn't exist? In this tree representation this can't happen since the folder has to first exist for an audience to even belong in it.

## Additionally, @janjelinek asked
### "What about empty folders?" Do you still show them even though all audiences inside it clearly only belong to Curated?

So I started thinking "ok, what's the data structure that would solve all the issues above, not show directories if they're empty, or if all audiences inside are of only one type, and also have it so we only compute it at the beginning?

To solve this I did the following:
from every audience we build a branch of the form
```
[ Branch audience1 [grandparent, parent, child]
, Branch audience2 []
, Branch audience3 [grandparent]
]
```
(grandparents are closest to root, and the empty list means we're at root)
For the sake of argument, let's say that all these audiences are of the same type.

Then, we walk through this list, take the folders, and construct nodes out of them, and then attach the audience at the right-most node.

This is perhaps clearer in json:

```json
{
    "grandparent" : {
        "audiences" : [],
        "subFolders" : {
            "parent" : {
                "audiences" : [],
                "subFolders" : {
                    "child" : {
                        "audiences" : [ { "title": "audience1"} ],
                        "subFolders" : {}
                    }
                }
            }
        }
    }
}
```

Then we pass this structure down to a function that takes the next branch and this tree, recursively descends into it, and inserts the audience at the right depth.
In this case, for the second branch, since it doesn't have any parent, we'd just insert that audience in `grandparent.audiences`, and that's it.

I tried to model this with the following type:
```elm
type AudienceTree =
    AudienceNode
        { id : Int
        , title : String
        , subFolders : Dict.Dict Int AudienceTree
        , audiences : List Data.Audience.Audience -- we could make this more general by putting "List a"
        }
```

We'd still need to walk through each branch to add it to the tree, so that would cost us O(n), and I chose Dicts because then digging inside this structure and updating it would cost us O(log n), resulting in the entire price of O(n log n) to construct the entire tree.

And in doing this, since we're starting from the audiences themselves, we wouldn't even get to directories that don't have audiences inside of them, thus solving the question @janjelinek asked about empty directories.

Now there would obviously be problems with this approach. 
For example, if the user can create directories, that information gets sent to the server, the user closes the browser window, opens it again, and then the user doesn't get shown the directory they've made because "it doesn't have any audience in it". 
But I mostly did this as an exercise for myself to try to come up with something that would solve all posed questions :D

## Disclaimers
What I wrote may be buggy. Unfortunately I did not take any measures to try to prevent this from happening, so whatever happens in the `Data.AudienceTree.attachBranch` may or may not be entirely correct, but it does produce output. In the real world I would obviously take care of something like this, but again, I mostly implemented this for myself just as an exercise.

We could've perhaps used something like a [rosetree](https://package.elm-lang.org/packages/zwilias/elm-rosetree/latest) with a [Zipper](https://package.elm-lang.org/packages/zwilias/elm-rosetree/latest/Tree-Zipper) to navigate the tree, but again, I wanted to try to write something myself as an exercise.

## Additional notes

If anything is uncertain, please don't hesitate to ping me either here or on slack @lawsdontapplytopigs!

PR #19, opened 2020-10-06
Challenge Submission

ursi:
I've finished the basic functionality for part one, but I haven't spent any time on making it look pretty, so my submission is not complete yet. However, what needs to be done is mostly additions, not modifications, so feel free to comment on what is already here.

To run the app you'll need to use [elm-git-install](https://github.com/robinheghan/elm-git-install). and then just fire up `elm reactor`.

Initially I'm going to use my own `elm-css` package (that is a superset of `elm/html`), as I find that nicer to work with. then, as per the rules, I can rewrite it using `elm/html` and CSS if need be.

### DIFF
--- a/elm-git.json
+++ b/elm-git.json
@@ -1,0 +1,8 @@
+{
+    "git-dependencies": {
+        "direct": {
+            "https://github.com/ursi/elm-css.git": "0.2.3"
+        },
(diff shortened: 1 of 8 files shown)


ursi:
I am not based in Europe :,(

PR #20, opened 2020-10-13
Question about Audience and AudienceFolder size

pogiaron:
Hi. I started to implement the challenge. Now I am thinking about performance, whether my solution would scale well.
In the fixtures there are 212 Audiences and 5 Folders.   
Can I know, in a real world scenario approximately how large the audiences, and folders would get?


Janiczek:
Hello @pogiaron,

> Can I know, in a real world scenario approximately how large the audiences, and folders would get?

* audiences = let's say 100-2000
* audience folders = let's say 5-100
It depends on how much the user uses this functionality.

If you can outline your thought process on what solutions there might be, and which one you chose and why, that would be great!
(Eg. "if it was on the order of 10000s of audiences and 1000s of folders, I'd choose XYZ instead because FOO...")

EDIT: oh and btw, super cool that you're doing the challenge! :)

pogiaron:
Hi @Janiczek, 
Thanks for answering my question, this is my story about the things I was thinking as I implemented the challenge.
### JSON parsing
For decoding I used [_elm-community/json-extra_](https://package.elm-lang.org/packages/elm-community/json-extra/latest/) . From this package, I only used the _andMap_ function to decode in an applicative style. Because the audiences and audience folders were not that big objects probably using the _mapN_ decoders would be good also.  

I was thinking about how could I decode the lists into dictionaries immediately where the keys would be the folders and the values would be the folder subfiles, and subfolders: `Dict Int (List Audience), Dict Int (List AudienceFolder)`
I came up with a solution with the help of the index decoder. I decoded a list element into a _Dict_ then I iterated over the list adding elements to the _Dict_, but I did not think it was pretty. I could build the dictionaries after decoding into lists, but then I would have to iterate the lists again, probably it would matter only if the lists would be really big.
### Modeling
At first, I came up with the simplest solution I could:

- Solution 0:

```
type alias AudiencesAndFolder =
    { audienceFolders : List AudienceFolder
    , audiences : List Audience
    , currentAudienceFolders : List AudienceFolder
    , currentAudiences : List Audience
    , currentFolder : Maybe AudienceFolder
    }
```
Always filtering the Audiences and AudienceFolders when the user would go to a folder or go up. 
There is a better solution using a stack. Pushing the current level into the stack when the user goes to a folder. Popping the stack when the user goes up. In this way when the user goes up it is not needed to search in folders and files. 

- Solution 1:

```
type FileSystem a b = FileSystem (List (CurrentLevel a b)) (Blocks a b)

type alias CurrentLevel a b = 
    { folder: Maybe a
    , subFolders_: List a
    , subFiles_: List b
    }

type alias Blocks a b =
    { folders : List a --Dict Int (List a)
    , files : List b --Dict Int (List b)
    }
```
With this solution when the user visits a folder again after he go up, the previously filtered subfiles, subfolders would still be lost.
So I implemented a multiway tree with a zipper, with the help of http://learnyouahaskell.com/zippers#taking-a-walk.
I wanted to implement it for myself to understand zippers better. 

- Solution 2:

```
type Explorer a b	
    = Root (List (Explorer a b))	
    | Folder a (List (Explorer a b))	
    | File b

type Crumb a b	
    = Crumb a (List (Explorer a b)) (List (Explorer a b))	
    | RootCrumb (List (Explorer a b)) (List (Explorer a b))

type Zipper a b	
    = Zipper (Explorer a b) (List (Crumb a b))
```
With this solution when we look up a folder's subfiles or it's subfolders, we don't lose this information. The multiway tree is more tune with the problem, but because only one folder should be shown it is not a must.
Then I rethought my solution a little bit and I came up with:

- Solution 3:

```
-- MTree
type MTree a
    = MNode a (List (MTree a))

-- Zipper
type Crumb a
    = Crumb a (List (MTree a)) (List (MTree a))

type Zipper a
    = Zipper (MTree a) (List (Crumb a))

-- FileSystem
type alias FileSystem a b =
    Zipper (FSEntry a b)

type FSEntry a b
    = Root
    | Folder a
    | File b
```
In this solution, I did not build the Folders, and Files into my multiway tree type. In this approach, I like that I get more generic/abstract datastructures: multiway tree and zipper.

I was also thinking about decoding into an array. Then only store array indexes in my data structure.

### Performance
To find a folder's subfiles and subfolders it is a must to iterate over the audiences and audience folders.
If audiences and audience folders would get large. Filtering the lists can become problematic. Building dicts from the lists can help. The time complexity of building the dicts cost O(n*log n) but finding subfolders subfiles would not be O(n) but O(log n).

In my solution, I build the multiway tree as the user opens folders for all folders I also store if it has been opened already. So when the user opens the same folder the subfiles subfolders do not need to be looked up again.   
Another solution can be to build the full tree at once beforehand. Building the full tree using dicts would cost O(n*log n). For each files I would have to find their subfolders/subfiles. With this solution, the user needs to wait longer until he sees the folders, but when he browses the system he did not have to wait.

With the 3. solution I could not store the folders and files separately, while this can be done with the 2. solution. With the 3. solution to get sub files and sub folders I have to iterate over a folder children twice.

Janiczek:
> In my solution, I build the multiway tree as the user opens folders for all folders I also store if it has been opened already. So when the user opens the same folder the subfiles subfolders do not need to be looked up again.

Wow, that sounds pretty much like the solution we're using in the production for Questions/Categories (different entities)! We have so many of these that preprocessing into a tree would freeze the UI for a few seconds; and so we have a lazy tree with zipper, doing the preprocessing only when it's needed and remembering it in case an user opens that folder again.

(But for audiences and audience folders we didn't go this way - the performance is good enough with a simple solution :) )

pogiaron:
I started to implement the second step. I noticed there is a curated property in the AudienceFolder fixtures.   
Should I separate the audience folders also?

Janiczek:
> Should I separate the audience folders also?

Yes you can! Curated folders to only be shown when the Curated filter is active.

Áron, I just want to warn you before doing further work that our hiring process for the vacant position is nearly finished as we're in talks with one candidate and basically almost done. You're of course free to finish the challenge if you want, and we will take a look (let me just say that writing a tree zipper from scratch is impressive!) -- but actual next steps in any hiring process (interviews, ...) would have to wait until our next position opens (which we expect to be around Q1-Q2 2021). So, if you were doing the challenge just for fun, then feel free to continue, but if you were doing it as a part of job application, then it might have to wait. We would like to get in touch with you as the position opens later (if you're still interested in 2021).

pogiaron:
Ah ok, thanks for the info. I wanted to apply for the positions but it seems it has to wait. Q1-Q2 can be good for me so we should definitely get in touch. I think I will finish the challenge soon, I would appreciate if you could review it if you have the time.

pogiaron:
I'm ready. Thanks for the challenge. I really enjoyed it.

Janiczek:
@pogiaron A colleague's team (at the same company) is searching for an Elm developer since today. I've notified him that you've applied to us but our position was already filled - if you would like to try to apply to his team, could you please send him an email at `dhnyk@globalwebindex.net`?

pogiaron:
Thanks, I sent him an email.

PR #21, opened 2020-10-16
[WIP] : Implementation of first and second part

nitrajka:
Hi,

I have finished the challenge functionally. However, I have some things to discuss.

1. Is using linter crucial in this challenge?
2. In the description of the second part, I could not find how folders should be filtered according to the categories. Therefore, I did not filter folders when displaying. Was there this kind of information?
3. This question is about Elm's best practice. I implemented also the second part of the challenge. Should I create a specific variable in `Model` to hold what I want to display and thus simplify view functions? Or is it okay to have some state and let the view function figure out what to display? I solved this as the latter question explains, but I do not think this is the best practise, since there is a logic in a view function. I do not think, there should be logic in a view function.

### Preview
Screen Shot 2020-10-16 at 18 31 38
Screen Shot 2020-10-16 at 18 31 57

Dark blue are clickable folders and light blue are files.

### How to run the project?
I was thinking about configuring webpack-dev-server, but to deliver the results fast, I simply used `elm reactor`. There was not needed any interaction with Javascript.

### Thigs I would to differently in production
I would organize CSS styles better as I think, that styled do not belong in HTML.

### DIFF
--- a/src/Data/Audience.elm
+++ b/src/Data/Audience.elm
@@ -1,6 +1,6 @@
 module Data.Audience exposing
     ( AudienceType(..), Audience
-    , audiencesJSON
+    , audiencesJSON, decodeAudiences
     )
 
 {-| Data.Audiences module
@@ -13,6 +13,8 @@ This module implements everything related to audience resource.
 @docs AudienceType, Audience, audienceJSON
 
 -}
+
+import Json.Decode as JD
 
 -- Type definition
 
@@ -34,7 +36,30 @@ type alias Audience =
     , folder : Maybe Int
     }
 
+{-| Decoder for an Audience
+-}
+decodeAudience : JD.Decoder Audience
+decodeAudience = JD.map4
+    Audience
+        (JD.field "id" JD.int)
+        (JD.field "name" JD.string)
+        (JD.field "type" decodeAudienceType)
+        (JD.field "folder" (JD.maybe JD.int))
 
+{-| Decoder for AudienceType
+-}
+decodeAudienceType : JD.Decoder AudienceType
+decodeAudienceType = JD.string |> JD.andThen (\str ->
+    case str of
+        "curated" -> JD.succeed Curated
+        "shared" -> JD.succeed Shared
+        "user" -> JD.succeed Authored
+        sthElse -> JD.fail <| "Unknown AudienceType: " ++ sthElse)
+
+{-| Decoder for List of Audiences
+-}
+decodeAudiences : JD.Decoder (List Audience)
+decodeAudiences = JD.field "data" (JD.list decodeAudience)
 
 -- Fixtures
 
--- a/src/Main.elm
+++ b/src/Main.elm
@@ -44,7 +44,6 @@ audiences : Dict Int (List Audience)
 audiences =
     audiencesJSON
     |> decodeString decodeAudiences
-    --|> log "here"
     |> Result.withDefault []
     |> map (\aud -> (M.withDefault -1 aud.folder, [aud]))
     |> DE.fromListDedupe List.append
@@ -73,8 +72,10 @@ displayContents model =
 
 maybeDisplayBack : Bool -> Html Msg
 maybeDisplayBack display =
-    if display then Html.button [Html.Events.onClick Up] [Html.text "Back"]
-    else Html.text ""
+    if display then
+        Html.button ((Html.Events.onClick Up)::backButtonStyle) [Html.text "Back"]
+    else
+        Html.div [] []
 
 displayFolder : AudienceFolder -> Html Msg
 displayFolder {id, name, parent} =
@@ -105,3 +106,21 @@ commonStyle = [ Html.Attributes.style "color" "#fff"
         , Html.Attributes.style "transition" "color .15s"
         , Html.Attributes.style "width" "100%"
     ]
+
+backButtonStyle = [ Html.Attributes.style "margin-left" "40px"
+    , Html.Attributes.style "background-color" "#7d30a5"
+    , Html.Attributes.style "cursor" "pointer"
+    , Html.Attributes.style "color" "#fff"
+    , Html.Attributes.style "border-color" "#007bff"
+    , Html.Attributes.style "display" "table"
+    , Html.Attributes.style "font-weight" "400"
+    , Html.Attributes.style "text-align" "center"
+    , Html.Attributes.style "white-space" "nowrap"
+    , Html.Attributes.style "vertical-align" "middle"
+    , Html.Attributes.style "user-select" "none"
+    , Html.Attributes.style "border" "1px solid transparent"
+    , Html.Attributes.style "padding" ".375rem .75rem"
+    , Html.Attributes.style "font-size" "1rem"
+    , Html.Attributes.style "line-height" "1.5"
+    , Html.Attributes.style "border-radius" ".25rem"
+    , Html.Attributes.style "transition" "color .15s"]
--- a/src/Main.elm
+++ b/src/Main.elm
@@ -7,7 +7,7 @@ import Html.Events
 import Html.Attributes
 import Browser exposing (sandbox)
 import Json.Decode exposing (decodeString)
-import Dict.Extra as DE
+import Dict.Extra as DE exposing (fromListDedupe)
 import Dict exposing (Dict, size)
 import List exposing (map, drop, filter)
 import Maybe as M exposing (withDefault)
@@ -65,27 +65,20 @@ viewContents : Model -> Html Msg
 viewContents model =
     let
         currF = List.head model.currentFolder |> withDefault -1
-    in
-    if model.audienceFilter /= Shared then
-        Html.div (contentsStyle (currF /= -1))
-        [maybeViewBackButton (currF /= -1)
-        , viewFilterButtons model.audienceFilter
-        , Html.ul [Html.Attributes.style "width" "50%"]
-            <| List.append
+        contents =
+            if model.audienceFilter /= Shared then
+                List.append
                 (viewAudienceFolders model.audienceFolders currF)
                 (viewAudiences model.audienceFilter model.audiences currF)
+            else
+                viewAllAudiences
+                <| Dict.foldl (\parent children lst -> children++lst) [] model.audiences
+    in
+    Html.div (contentsStyle (currF /= -1))
+        [ maybeViewBackButton (currF /= -1 && model.audienceFilter /= Shared)
+        , viewFilterButtons model.audienceFilter
+        , Html.ul [Html.Attributes.style "width" "50%"] contents
         ]
-    else
-        Html.div (contentsStyle (currF /= -1))
-            [viewFilterButtons model.audienceFilter
-            , Html.ul [Html.Attributes.style "width" "50%"]
-                (viewAllAudiences
-                    <| Dict.foldl (\parent contents lst -> contents++lst) [] model.audiences)
-            ]
-
-viewAllAudiences : List Audience -> List (Html Msg)
-viewAllAudiences ads =
-    map (\file -> viewAudience file) ads
 
 maybeViewBackButton : Bool -> Html Msg
 maybeViewBackButton display =
@@ -96,26 +89,32 @@ maybeViewBackButton display =
 
 viewFilterButtons : AudienceType -> Html Msg
 viewFilterButtons at =
+    Html.div
+        [ Html.Attributes.style "width" "50%"
+        , Html.Attributes.style "margin-left" "40px"
+        ]
+        [ viewFilterButton at Authored
+        , viewFilterButton at Shared
+        , viewFilterButton at Curated
+        ]
+
+viewFilterButton : AudienceType -> AudienceType -> Html Msg
+viewFilterButton selectedAudienceType audienceType =
     let
         filterButtonStyles = containerStyle++containerButtonStyle
     in
-    Html.div
-        [Html.Attributes.style "width" "50%"
-        , Html.Attributes.style "margin-left" "40px"
-        ]
-        [ Html.button
-            (filterButtonStyles++(maybeSelectedAudience (at == Authored) (Filter Authored)))
-            [text "Authored"]
-        , Html.button
-            (filterButtonStyles++(maybeSelectedAudience (at == Shared) (Filter Shared)))
-            [text "Shared"]
-        , Html.button
-            (filterButtonStyles++(maybeSelectedAudience (at == Curated) (Filter Curated)))
-            [text "Curated"]
-        ]
+    Html.button
+        (filterButtonStyles++
+            (selectedAudienceAction (selectedAudienceType == audienceType) (Filter audienceType))
+        )
+        [text <| toString audienceType]
 
-maybeSelectedAudience : Bool -> Msg -> List (Html.Attribute Msg)
-maybeSelectedAudience isSelected msg =
+viewAllAudiences : List Audience -> List (Html Msg)
+viewAllAudiences ads =
+    map (\file -> viewAudience file) ads
+
+selectedAudienceAction : Bool -> Msg -> List (Html.Attribute Msg)
+selectedAudienceAction isSelected msg =
     let
         action = [Html.Events.onClick msg]
     in
@@ -152,6 +151,10 @@ viewAudience {id, name, type_, folder} =
             [ Html.Attributes.style "background-color" "#91c0f3"
             , Html.Attributes.style "border-color" "#8cacce" ])
         [Html.text name]
+
+
+----- styles only ----
+
 
 contentsStyle : Bool -> List (Html.Attribute Msg)
 contentsStyle backExists =
--- a/src/Main.elm
+++ b/src/Main.elm
@@ -63,42 +63,33 @@ displayContents model =
     let
         currF = List.head model.currentFolder |> withDefault -1
     in
-    --[Html.text <| toString (size model.audiences), Html.text <| toString (size model.audienceFolders)]
-    Html.ul [ Html.Attributes.style "width" "50%"]
-    <| List.append
-        (map displayFolder (Dict.get currF model.audienceFolders |> withDefault []))
-        (map displayFile (Dict.get currF model.audiences |> withDefault []))
+    Html.div []
+    [maybeDisplayBack (currF /= -1)
+    , Html.ul [ Html.Attributes.style "width" "50%"]
+        <| List.append
+            (map displayFolder (Dict.get currF model.audienceFolders |> withDefault []))
+            (map displayFile (Dict.get currF model.audiences |> withDefault []))
+    ]
 
+maybeDisplayBack : Bool -> Html Msg
+maybeDisplayBack display =
+    if display then Html.button [Html.Events.onClick Up] [Html.text "Back"]
+    else Html.text ""
 
 displayFolder : AudienceFolder -> Html Msg
 displayFolder {id, name, parent} =
-    Html.li
-        [ Html.Attributes.style "color" "#fff"
+    Html.li (commonStyle++
+        [ Html.Events.onClick (Down id)
         , Html.Attributes.style "background-color" "#007bff"
         , Html.Attributes.style "cursor" "pointer"
-        , Html.Attributes.style "border-color" "#007bff"
-        , Html.Attributes.style "display" "table"
-        , Html.Attributes.style "font-weight" "400"
-        , Html.Attributes.style "text-align" "center"
-        , Html.Attributes.style "white-space" "nowrap"
-        , Html.Attributes.style "vertical-align" "middle"
-        , Html.Attributes.style "user-select" "none"
-        , Html.Attributes.style "border" "1px solid transparent"
-        , Html.Attributes.style "padding" ".375rem .75rem"
-        , Html.Attributes.style "font-size" "1rem"
-        , Html.Attributes.style "line-height" "1.5"
-        , Html.Attributes.style "border-radius" ".25rem"
-        , Html.Attributes.style "transition" "color .15s"
-        , Html.Attributes.style "width" "100%"
-        , Html.Events.onClick (Down id)
-        ]
+        ])
         [ Html.text name ]
 
 displayFile : Audience -> Html Msg
 displayFile {id, name, type_, folder} =
-    Html.li [ Html.Attributes.style "color" "#fff"
-        , Html.Attributes.style "background-color" "#91c0f3"
-        , Html.Attributes.style "cursor" "pointer"
+    Html.li ((Html.Attributes.style "background-color" "#91c0f3")::commonStyle) [Html.text name]
+
+commonStyle = [ Html.Attributes.style "color" "#fff"
         , Html.Attributes.style "border-color" "#007bff"
         , Html.Attributes.style "display" "table"
         , Html.Attributes.style "font-weight" "400"
@@ -113,4 +104,4 @@ displayFile {id, name, type_, folder} =
         , Html.Attributes.style "border-radius" ".25rem"
         , Html.Attributes.style "transition" "color .15s"
         , Html.Attributes.style "width" "100%"
-    ] [Html.text name]
+    ]
--- a/src/Main.elm
+++ b/src/Main.elm
@@ -1,10 +1,116 @@
 module Main exposing (main)
 
-import Data.Audience
-import Data.AudienceFolder
-import Html exposing (Html)
+import Data.Audience exposing (..)
+import Data.AudienceFolder exposing (..)
+import Html exposing (Html, div, button, text)
+import Html.Events
+import Html.Attributes
+import Browser exposing (sandbox)
+import Json.Decode exposing (decodeString)
+import Dict.Extra as DE
+import Dict exposing (Dict, size)
+import List exposing (map, drop)
+import Maybe as M exposing (withDefault)
+import Debug exposing (toString)
 
 
-main : Html msg
-main =
-    Html.text "There will be app soon!"
+main : Program () Model Msg
+main = sandbox {
+    init = {audiences = audiences, audienceFolders = audienceFolders, currentFolder = []},
+    view = view,
+    update = update}
+
+type alias Model = {
+    audiences : Dict Int (List Audience)
+    , audienceFolders :  Dict Int (List AudienceFolder)
+    , currentFolder : List Int }
+
+type Msg =
+    Up
+    | Down Int
+
+view : Model -> Html Msg
+view model =
+    div [] [displayContents model]
+
+update : Msg -> Model -> Model
+update msg model =
+    case msg of
+        Up -> {model | currentFolder = drop 1 model.currentFolder}
+        Down dir -> {model | currentFolder = dir::model.currentFolder}
+
+
+audiences : Dict Int (List Audience)
+audiences =
+    audiencesJSON
+    |> decodeString decodeAudiences
+    --|> log "here"
+    |> Result.withDefault []
+    |> map (\aud -> (M.withDefault -1 aud.folder, [aud]))
+    |> DE.fromListDedupe List.append
+
+audienceFolders : Dict Int (List AudienceFolder)
+audienceFolders =
+    audienceFoldersJSON
+    |> decodeString decodeAudienceFolders
+    |> Result.withDefault []
+    |> map (\audF -> (M.withDefault -1 audF.parent, [audF]))
+    |> DE.fromListDedupe List.append
+
+
+displayContents : Model -> Html Msg
+displayContents model =
+    let
+        currF = List.head model.currentFolder |> withDefault -1
+    in
+    --[Html.text <| toString (size model.audiences), Html.text <| toString (size model.audienceFolders)]
+    Html.ul [ Html.Attributes.style "width" "50%"]
+    <| List.append
+        (map displayFolder (Dict.get currF model.audienceFolders |> withDefault []))
+        (map displayFile (Dict.get currF model.audiences |> withDefault []))
+
+
+displayFolder : AudienceFolder -> Html Msg
+displayFolder {id, name, parent} =
+    Html.li
+        [ Html.Attributes.style "color" "#fff"
+        , Html.Attributes.style "background-color" "#007bff"
+        , Html.Attributes.style "cursor" "pointer"
+        , Html.Attributes.style "border-color" "#007bff"
+        , Html.Attributes.style "display" "table"
+        , Html.Attributes.style "font-weight" "400"
+        , Html.Attributes.style "text-align" "center"
+        , Html.Attributes.style "white-space" "nowrap"
+        , Html.Attributes.style "vertical-align" "middle"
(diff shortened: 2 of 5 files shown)


Janiczek:
Hi and thanks for the solved challenge! We'll take a look and review it.

> 1. Is using linter crucial in this challenge?

No it is not, but you're free to do so! (polish the project in whatever amount you want)

----

> 2. In the description of the second part, I could not find how folders should be filtered according to the categories. Therefore, I did not filter folders when displaying. Was there this kind of information?

Thanks for pointing this spec omission out. There is just one mention, about shared audiences:

> In case of shared audiences ignore tree structure and display audiences in single level without any folder.

So that one will not have folders at all. The other ones will have folders, but perhaps you can hide a folder if it contains no audiences of the currently filtered type. Eg.

```
audiences:
AA (authored, shared) inside folder FA
AB (default) inside folder FB
AC (shared) inside folder FA
AD (shared) inside folder FD
AE (authored) inside folder FE
```
would let you see this **in root**:
* When looking at Authored audiences: you see folders FA, FE
* When looking at Shared audiences: you see no folders (but you see 3 audiences)
* When looking at Default audiences: you see folder FB

----

> 3. This question is about Elm's best practice. I implemented also the second part of the challenge. Should I create a specific variable in Model to hold what I want to display and thus simplify view functions? Or is it okay to have some state and let the view function figure out what to display? I solved this as the latter question explains, but I do not think this is the best practise, since there is a logic in a view function. I do not think, there should be logic in a view function.

This is debatable. Basically it's about "recompute all the time, or cache the results?" Caching is more performant (less work done overall) but has the danger of leaving the cache stale and showing wrong results. That's the benefit of keeping the logic in `view` - the results are always fresh.

I believe the default should be: just leave the filtering/... logic in `view` and if performance is an issue (often it's not), first try tweaking it with `Html.Lazy` (saves you some "child" `fooView` calls) or `Html.Keyed`. If that doesn't help, then perhaps it's time to start thinking about a cache. We have such a cache in a few places but it's generally more complex and less maintainable code, so we try to do stuff in `view` whenever we can.

@janjelinek @kraklin @annaghi thoughts?

kraklin:
> We have such a cache in a few places but it's generally more complex and less maintainable code, so we try to do stuff in view whenever we can. 

Well, I'd say as always in IT, "It depends". But I think we are not trying to do the stuff in view "whenever" we can. As there are places where we try to do stuff in update whenever we can (looking your way, XB.TableView) :D 

But in general I do agree with the steps like `view` -> `Html.Lazy` -> `cache/update`.

src/Data/Audience.elm
```diff
@@ -34,7 +36,30 @@ type alias Audience =
     , folder : Maybe Int
     }
 
+{-| Decoder for an Audience
+-}
+decodeAudience : JD.Decoder Audience
+decodeAudience = JD.map4
+    Audience
```
Janiczek, line 43: Just a nitpick remark - I was quite confused by this the first time I saw it - I assumed it's `elm-format` processed and so it didn't make sense to me how can `Audience ...` be a decoder -- but then I saw the `JD.map4` on the previous line :smile:

(We use elm-format everywhere in our codebase - and I can feel the benefits of having it right now on this anecdote :) less surprises when reading "foreign" code...)

src/Data/Audience.elm
```diff
@@ -34,7 +36,30 @@ type alias Audience =
...
+        (JD.field "id" JD.int)
+        (JD.field "name" JD.string)
+        (JD.field "type" decodeAudienceType)
+        (JD.field "folder" (JD.maybe JD.int))
```
Janiczek, line 47: If there were more than 8 fields in an Audience to decode (so that you couldn't use `JD.map8`-like approach), what decoding approach would you choose?

src/Main.elm
```diff
@@ -1,10 +1,198 @@
...
+import Maybe as M
 
 
-main : Html msg
-main =
-    Html.text "There will be app soon!"
+type alias Model = {
+    audiences : Dict Int (List Audience)
+    , audienceFolders :  Dict Int (List AudienceFolder)
+    , currentFolder : List Int
```
Janiczek, line 19: It would be nice to have somehow documented the order of things that live in this `currentFolder` field (is it newest first, or oldest first?)

Perhaps even using a different data structure would make it more obvious (`: Stack Int` - stacks are LIFO) but it might be an overkill so I'd be fine with `-- newest first`-like comment.

BTW the name of the field could be more descriptive - something like `breadcrumbs`? `pathToRoot`? `currentFolder` suggests there's just one item but in fact it's a list...

src/Main.elm
```diff
@@ -1,10 +1,198 @@
...
+view : Model -> Html Msg
+view model =
+    div [] [viewContents model]
+
+update : Msg -> Model -> Model
+update msg model =
+    case msg of
+        Up -> {model | currentFolder = L.drop 1 model.currentFolder}
+        Down dir -> {model | currentFolder = dir::model.currentFolder}
+        Filter at -> {model | audienceFilter = at, currentFolder = []}
```
Janiczek, line 45: nice and succint :+1:

src/Main.elm
```diff
@@ -1,10 +1,198 @@
...
+
+audiences : Dict Int (List Audience)
+audiences =
+    audiencesJSON
+    |> JD.decodeString decodeAudiences
+    |> Result.withDefault []
```
Janiczek, line 51: Acceptable for the purposes of the challenge :+1:
Can you just sketch what would you probably do if you had to `GET` these JSONs in runtime? eg. what would be the types in the `Model`, etc.

src/Main.elm
```diff
@@ -1,10 +1,198 @@
...
+    |> L.map (\aud -> (M.withDefault -1 aud.folder, [aud]))
+    |> DE.fromListDedupe List.append
```
Janiczek, line 53: First time I see [this function](https://package.elm-lang.org/packages/elm-community/dict-extra/latest/Dict-Extra#fromListDedupe), thanks for making me aware of it!

src/Main.elm
```diff
@@ -1,10 +1,198 @@
...
+        Up -> {model | currentFolder = L.drop 1 model.currentFolder}
+        Down dir -> {model | currentFolder = dir::model.currentFolder}
+        Filter at -> {model | audienceFilter = at, currentFolder = []}
+
+audiences : Dict Int (List Audience)
+audiences =
+    audiencesJSON
+    |> JD.decodeString decodeAudiences
+    |> Result.withDefault []
+    |> L.map (\aud -> (M.withDefault -1 aud.folder, [aud]))
```
Janiczek, line 52: The way I understand it, `-1` is the key of the root level in this Dict? So the dict is from parent ID to list of children?

EDIT: why I'm commenting on this - normally I'd expect it to be a dict from audience ID to the audience definition, or from folder ID to folder definition. This is more of a preprocessing step for the folder structure, it seems. Interesting!

src/Main.elm
```diff
@@ -1,10 +1,198 @@
...
+            if model.audienceFilter /= Shared then
+                List.append
+                (viewAudienceFolders model.audienceFolders currF)
+                (viewAudiences model.audienceFilter model.audiences currF)
+            else
+                viewAllAudiences
+                <| Dict.foldl (\parent children lst -> children++lst) [] model.audiences
+    in
+    Html.div (contentsStyle (currF /= -1))
+        [ maybeViewBackButton (currF /= -1 && model.audienceFilter /= Shared)
```
Janiczek, line 77: I have to say I don't like the `-1` magic number much.
Elm allows us to use meaningful types like
```elm
type Position
  = Root -- instead of -1
  | Folder Int
```
So I'd prefer that

src/Main.elm
```diff
@@ -1,10 +1,198 @@
...
+    audienceFoldersJSON
+    |> JD.decodeString decodeAudienceFolders
+    |> Result.withDefault []
+    |> L.map (\audF -> (M.withDefault -1 audF.parent, [audF]))
+    |> DE.fromListDedupe List.append
+
+viewContents : Model -> Html Msg
+viewContents model =
+    let
+        currF = List.head model.currentFolder |> M.withDefault -1
```
kraklin, line 66: I like more descriptive names. `currF` seems too cryptic to me. Also I'm not sure what does the `-1` means

src/Main.elm
```diff
@@ -1,10 +1,198 @@
...
+        contents =
+            if model.audienceFilter /= Shared then
+                List.append
+                (viewAudienceFolders model.audienceFolders currF)
+                (viewAudiences model.audienceFilter model.audiences currF)
+            else
+                viewAllAudiences
+                <| Dict.foldl (\parent children lst -> children++lst) [] model.audiences
```
kraklin, line 74: `lst` should mean `last` here? I'm more used to accumulator or `acc` :)

src/Main.elm
```diff
@@ -1,10 +1,198 @@
...
 
-import Data.Audience
-import Data.AudienceFolder
-import Html exposing (Html)
+import Browser exposing (sandbox)
+import Data.Audience exposing (..)
+import Data.AudienceFolder exposing (..)
+import Dict exposing (Dict)
+import Dict.Extra as DE exposing (fromListDedupe)
+import Html exposing (Html, div, button, text)
```
kraklin, line 8: out of curiosity - why did you choose to expose functions like `div`, `button` etc, but later you are using `Html.Attributes.style` quite a lot. Also you have aliases for `Json.Decode` as `D` but none for `Html.Attributes` nor `Html.Events`. Was this the deliberate choice?

src/Main.elm
```diff
@@ -1,10 +1,198 @@
...
+                <| Dict.foldl (\parent children lst -> children++lst) [] model.audiences
+    in
+    Html.div (contentsStyle (currF /= -1))
+        [ maybeViewBackButton (currF /= -1 && model.audienceFilter /= Shared)
+        , viewFilterButtons model.audienceFilter
+        , Html.ul [Html.Attributes.style "width" "50%"] contents
+        ]
+
+maybeViewBackButton : Bool -> Html Msg
+maybeViewBackButton display =
```
kraklin, line 83: `maybeView` implies here that the function is deciding on `Maybe`, I would not expect `Bool` to be there from the function name.

But maybe I'm too accustomed to our codebase where we use `maybe` prefix just for that. I'd choose something like `showBackButton` or just make it `viewBackButton`. Naming is hard :)

src/Main.elm
```diff
@@ -1,10 +1,198 @@
...
+viewAllAudiences : List Audience -> List (Html Msg)
+viewAllAudiences ads =
+    L.map (\file -> viewAudience file) ads
+
+viewAudiences : AudienceType -> Dict Int (List Audience) -> Int -> List (Html Msg)
+viewAudiences at ads currentFolder =
+    L.map viewAudience (L.filter (\a -> a.type_ == at) (Dict.get currentFolder ads |> M.withDefault []))
+
+viewAudience : Audience -> Html Msg
+viewAudience {id, name, type_, folder} =
```
kraklin, line 152: `type_` is redundant here as it is not used in the body of the function

kraklin (commented):
Hey @nitrajka 

thanks for taking your part in this challenge. I hope you liked it.

The code is quite nice and not extensively long and it is doing what it should do. I realise how I'm used to `elm-format` formatting as some parts were a little bit harder for me to read than usual.

I left few comments about the code, but overall it looks good to me. One questionable thing would be making those operations in the view instead in update, but that's probably not needed in the code like this.

Anyway, nice work.

src/Main.elm
Janiczek: perhaps `lst` means `list` :hand_over_mouth:

kraklin: 🤦‍♂️ that makes sense :) Also it shows how those non-descriptive names are hard to work with :D

Janiczek: To be fair, I think you could argue the same way about our `acc`. It's all a matter of convention in the team. Other teams perhaps are used to `lst`.

kraklin: Yes, that's true, still `lst` as in `list` can't be used when `accumulator` is e.g. `dict` or `array`. You can use `acc` all the time :D 

But point taken. 👍

src/Data/Audience.elm
```diff
@@ -34,7 +36,30 @@ type alias Audience =
...
+decodeAudienceType : JD.Decoder AudienceType
+decodeAudienceType = JD.string |> JD.andThen (\str ->
+    case str of
+        "curated" -> JD.succeed Curated
+        "shared" -> JD.succeed Shared
+        "user" -> JD.succeed Authored
+        sthElse -> JD.fail <| "Unknown AudienceType: " ++ sthElse)
+
+{-| Decoder for List of Audiences
+-}
```
janjelinek, line 60: This kind of comment is kinda redundant. 
You can just rename `decodeAudiences` to `audiencesDecoder` and than you do not need comment at all :)

src/Data/Audience.elm
nitrajka: I am sorry, I should have definitely used `elm-format`.

kraklin: Never mind. It is always good to be aware of the things you are addicted on using :D E.g. how much you miss them when they are not used ;)

src/Main.elm
```diff
@@ -1,10 +1,198 @@
...
-    Html.text "There will be app soon!"
+type alias Model = {
+    audiences : Dict Int (List Audience)
+    , audienceFolders :  Dict Int (List AudienceFolder)
+    , currentFolder : List Int
+    , audienceFilter : AudienceType }
+
+type Msg =
+    Up
+    | Down Int
```
janjelinek, line 24: Not much descriptive to be honest. In this case it's fine, but from "this will be maintained by various people in future" point of view I would rather choose something more clear for the first look.
Like `OpenFolder AudienceFolderId` and `BackFromFolder` or anything else. `Int` also did not telling much about it's real type (folderId).

src/Main.elm
janjelinek: Agree, interesting function.

src/Main.elm
janjelinek: Agree, also this has potential to be buggy when you will somehow get something like `[ -1, 123, -1, 444 ]` so you can possibly ended up in folder structure without chance to going back. This is in general weak point of whole approach of using `List Int` for navigation. It's OK with current code, but code is usually changed over the time and as we know, if something can break it will break :)

janjelinek (commented):
First, thanks for you submission and effort you put into it.
Most of the points were already said by others anyway I had to again mention `elm-format` :) 
It's totally ok for purposes of challenge to do not use it, but honestly reading code with inconsistent formatting is much harder. 

**Code style**
I also noticed inconsistency in approach. You had some redundant comments for obvious code, but also bunch of cryptic shortcuts for variable names without any comment, this is definitely space for you improvement, don't be afraid of longer names for variables and functions. Also I miss some more thinking about edge-cases and impossible states.

**CSS and styling**
I can see only minimal styling here without any external (S)CSS file and even this small piece of styles can't tell much about your abilities. You did not payed that much attention to it and you are aware of it (according to your PR description), but in the end we are working on FE 🤷 

**Summary**
I don't know your ELM experiences level, so for "I did not have previous production Elm experiences" it's quite nice and good code, but with clearly juniority here. But you know, there is always room for improvement so 👍

janjelinek:
Also you probably forgot remove `[WIP]` from PR name, right @nitrajka? Using drafts PR is for this purpose much better ;)

src/Data/Audience.elm
nitrajka: There are multiple ways how to decode a JSON object without using `mapN` in Elm. Basically, a pipeline and applicatives are a general idea here. 

```
apply : JD.Decoder a -> JD.Decoder (a -> b) -> JD.Decoder b
apply da df = JD.andThen (\f -> JD.map f da ) df

apply1 : JD.Decoder a -> JD.Decoder (a -> b) -> JD.Decoder b
apply1 da df = JD.map2 (\arg func -> func arg) da df

apply2 : JD.Decoder a -> JD.Decoder (a -> b) -> JD.Decoder b
apply2 da df = JD.map2 (|>) da df

apply3 : JD.Decoder a -> JD.Decoder (a -> b) -> JD.Decoder b
apply3 = JD.map2 (|>)

decodeAudience : JD.Decoder Audience
decodeAudience = JD.succeed Audience
    |> apply (JD.field "id" JD.int)
    |> apply1 (JD.field "name" JD.string)
    |> apply2 (JD.field "type" decodeAudienceType)
    |> apply3 (JD.field "folder" (JD.maybe JD.int))
```
JD.succeed Audience produces a value of type `Decoder a` (where, correct me if I am wrong, `a` is of type `c->d->e->f-> Audience` because Audience needs 4 parameters to be created). This value of type `Decoder a` is sent down the pipeline by operator `|>`. 
All of the apply functions are just Elm's version of `(<*>)` in Haskell (`(<*>)` is called bind for applicative in Haskell). 

The first argument for apply{ ,1,2,3} is always a (JD.Decoder a) value (for example `(JD.field "id" JD.int)`). The second argument for `apply` is a JD.Decoder (a->b) (this argument goes down the pipeline) which is `c->d->e->f-> Audience` for `apply`, `d->e->f-> Audience` for `apply1`, `e->f-> Audience` for `apply2` and `f-> Audience` for `apply3`.

Comment to `apply` function:
Function `andThen : (a -> Decoder b) -> Decoder a -> Decoder b` helps unpack the `df` function (i.e., df is of type `Decoder (c->d->e->f-> Audience)`) and so we can apply the unpacked function `f` (of type `c->d->e->f-> Audience`) to the da value (of type `Decoder a`) using simple `map : (a -> value) -> Decoder a -> Decoder value`. After applying map, the result is `Decoder (d->e->f->Audience)` which is sent to the next part in the pipeline.

Comment to `apply1` function:
Function `map2 : (a -> b -> value) -> Decoder a -> Decoder b -> Decoder value` takes as the first argument a function which takes "unpacked" values of types a and b and return some value. The second and third argument of `map2` are "packed" . `map2` unpacks those two arguments and applies a given function on the unpacked values. Then it packs the result (packing of applicatives may be done through `pure` in Haskell)

Comment to `apply2` function:
since we know, that in `apply1`, the first argument of `map2` is a function of type `a -> (a->b) -> b` we can safely use `(|>)` operator, instead of our custom function. 

Comment to `apply3` function:
We can commit the arguments since they are applied to `|>` in the same order as function `apply3` receives them.

I hope my answer was clear enough. If not, please, ask and let me explain.

I found one more neat solution: https://package.elm-lang.org/packages/webbhuset/elm-json-decode/latest/
The thing I like about this solution is that the record field order does not matter. 

Note:
My teacher (of Advanced functional programming) used terms "packed" and "unpack" to better explain, how functors, applicatives and monads work. That they are only values packed in some box and everytime we want to apply a packed function to a packed value, we need to unpack them first. Those terms helped me, therefore I use them. I hope you know what I mean.

I hope to write the rest of the comments until the end of the day. Sorry for the delay.

Janiczek: Thanks for the extensive comment, yeah applicatives are the way to go.
We're using JDE.andMap pattern at work.
```elm
Json.Decode.succeed Record
 |> Json.Decode.Extra.andMap (Json.Decode.field ...)
 |> Json.Decode.Extra.andMap (Json.Decode.field ...)
 |> Json.Decode.Extra.andMap (Json.Decode.field ...)
 |> Json.Decode.Extra.andMap (Json.Decode.field ...)
 |> Json.Decode.Extra.andMap (Json.Decode.field ...)
 |> Json.Decode.Extra.andMap (Json.Decode.field ...)
 |> Json.Decode.Extra.andMap (Json.Decode.field ...)
 |> Json.Decode.Extra.andMap (Json.Decode.field ...)
 |> Json.Decode.Extra.andMap (Json.Decode.field ...)
 |> Json.Decode.Extra.andMap (Json.Decode.field ...)
```

> I found one more neat solution: https://package.elm-lang.org/packages/webbhuset/elm-json-decode/latest/
The thing I like about this solution is that the record field order does not matter.

I wish the `andThen` lambda nesting wasn't formatted so badly by `elm-format`. ie. [this ellie](https://ellie-app.com/bjC2b4nxckma1)

Janiczek:
Thanks for the solved challenge!

Overall you obviously know Elm and the more advanced FP patterns. Our team honestly tends to stay on the more "obvious" / simple side of things most of the time, but there are a few folks in the company who have a more academic background and like to use the more advanced patterns... talking about you, @mathiajusth :smile: 

Anyways, the solution is good! I'm still wondering about your decoding into `Dict parentId (List childItem)` instead of `Dict childId childItem`. Something to think about for me! It makes sense as an optimization, if that is ever needed (probably not right from the start).

I'm pretty satisfied with this overall.

----

One more extra bonus question, if you want to tackle it :) If there were too much data (too many entities) to process, such that running that preprocessing step you did would take ~3s and freeze the UI thread, what different approach would you try?

(We have hit and solved that problem with a different set of entities, so it's not entirely made-up / fabricated problem.)

src/Main.elm
nitrajka: You are right, it should be obvious whether it is newest first or oldest first. I used the newest first because prepending in Haskell is cheaper than appending. Therefore I thought, that in Elm too. And the name of the field could be more descriptive, I agree.

Janiczek: Yeah! To be fair, I think it would be safe to bet that the list will grow/shrink at the beginning just because of how linked lists work.

Still, I think that when getting into the code after 2 months it's nice to be able to get up to speed with a well placed comment / type name / ... than having to Debug.log it to be sure or to look at how it's updated in `update`. :slightly_smiling_face:

src/Main.elm
nitrajka: Yes, you understand it well. `-1` is the key for root level (I know, not a good practise). And yes, the Dict is from parent ID to list of children. As you said, I preprocessed it to move up and down in the filesystem more quickly. Another option I was thinking about was custom zipper, but that would be slower than this solution. Each time I wanted to go down in zipper, I had to find the exact folder in the current node's list of folders, and this operation takes O(n) time. This is avoidable with Dict.

src/Main.elm
nitrajka: I agree. However, I knew the data in advance and database systems  (usually) start assigning IDs from 1. I was thinking about using `Maybe` but could not somehow get used to the idea, that I would be always `case`ing the type.
```
case head pathToRoot of 
    Just folder -> display folder's data
   Nothing -> displayRoot data
```
@Janiczek 's idea has the same "issue" with `case`ing, but at least it is clear that something is root and something is just Folder. However, a new case would arise => we would have to check whether the `pathToRoot` list is not empty.

src/Main.elm
nitrajka: Yes, you are right, it should have been more descriptive.
The `-1` is the specification for the root folder. I have just explained it in another comment in this review. I know, it is not a good practice.

src/Main.elm
nitrajka: Yes, `lst` means list here. If `acc` was used, one could think that accumulator can be anything that works with `++` operator (I am not sure now, how many options we have here). Therefore I wanted to make sure, that I am using a list as an accumulator.

src/Main.elm
nitrajka: From `Html` I exposed only `Html, button, div, text` because I used only those (did not want to import the whole library). As for the `Html.Attributes.style`, I just forgot to import only `style`. Yes, it is a bit inconsistent, sorry. In the next update, I will expose only what I need (if the line, where I expose thing is not very long :D ).

src/Main.elm
nitrajka: Well, every `if` statement is deciding on a bool value. Yes, it is not a `Maybe`, but still it can have only 2 possible outcomes. The function is deciding on a bool value whether to display the button or not.

src/Main.elm
nitrajka: Yes, you are right.

src/Data/Audience.elm
nitrajka: That is a good point :) thank you.

src/Main.elm
nitrajka: Yes, you are right. The action names should have been more descriptive.

src/Main.elm
kraklin: Well, in our codebase we use the full qualified names and don't import bare functions but only types. And for long module names we aliase them e.g.:

```elm
Html.div [Attrs.style "font-weight" "bold"] [Html.p [] [Html.text "wow, what a nice unicorn."] ]
```

this way we don't have the collisions when the names of functions in different modules are the same and we know exactly from where the function comes from without the need to look at the imports.

I think either way is fine

src/Main.elm
kraklin: I believe that only other option would be `String` and that it is a List can be understood from the `[]` after the `foldl`. Also I prefer `acc` as when you change the type, you don't have to change the name of variable (we don't need to have Hungarian notation anymore)

src/Main.elm
kraklin: Sorry for badly phrased question. It was about the name of the function implying the wrong input type. As the function name starts with `maybe` I expect `Maybe` to be decided on.

I would change the name, not the inner `if` nor `Bool` as an argument. Something like [viewIf](https://package.elm-lang.org/packages/elm-community/html-extra/3.4.0/Html-Extra#viewIf)

src/Main.elm
nitrajka: Good idea 👍 What you suggested seems more clear.
The way how one imports a function does not impact a performance (yes, if you import the whole library with 100000+ functions, then yes). As there are (correct me if I am wrong) more ways to import a function in Elm, Elm developers may argue which way of importing a function is better and this may create a tension in a team if all of the members are too much enthusiastic and each of them prefers his own coding style. :D If the syntax for importing would be restricted to fewer options (or at least a linter could handle it) then so Elm developers could solve more of the more serious coding issues and not discuss coding styles so often. 
I always had a feeling while doing frontends in Elm, or Javascript, that developers discussed more coding styles than actual algorithms or more technically challenging issues. But, maybe it is only my experience/feeling.

src/Main.elm
nitrajka: I think I understood your question. But I did not know, there is a convention. That when a function is deciding on `Maybe`, the function's name should start with `maybe...`. And when deciding on bool, the function's name should end with `If` as in the `viewIf` example.

src/Main.elm
Janiczek: > I always had a feeling while doing frontends in Elm, or Javascript, that developers discussed more coding styles than actual algorithms or more technically challenging issues. But, maybe it is only my experience/feeling.

Certainly true. But honestly, there aren't as many actual algorithms to solve stuff with on the frontend side of things. You need to present stuff to user, let user act on it, etc., but there isn't much of processing etc. - you get all that mostly handed to you on a silver plate by the API.

There still are interesting things and optimizations, but overall I can't expect FE programming to be as algorithmically challenging as eg. data processing.

> If the syntax for importing would be restricted to fewer options (or at least a linter could handle it) then so Elm developers could solve more of the more serious coding issues and not discuss coding styles so often.

`elm-format` removes some of that but there still are some degrees of freedom where folks clash and have to come to some kind of compromise if they work on a shared codebase.

nitrajka:
@Janiczek to the example which you described in the comment.
1. I thought that a single audience can be of a single type (either Shared, Authored or Curated). Not that an audience can be of multiple types. I am not sure, but I think, this information is not mentioned in here: https://github.com/GlobalWebIndex/challenge-elm/blob/master/SECOND_STEP.md
Also, I did not realize, I should not display folders which do not have any content (because of filtration). Was that specifically written to not display such folders?
Thanks for your answers. 

> One more extra bonus question, if you want to tackle it :) If there were too much data (too many entities) to process, such that running that preprocessing step you did would take ~3s and freeze the UI thread, what different approach would you try?

Unfortunately, I have never had to solve such a problem in Elm. However, I would suggest to fetch only changed objects from a backend. Of course, the first fetch would always be slow. Let's say, we would be able to identify the objects based on unique identifier which does not change (id). The first request on a backend would fetch all of the objects witch their ids. Each subsequent request on backend would return only ids of changed objects. After receiving ids of changed objects, only those  would be fetched from backend. I think, this solution would perform better in the server-client communication, however, it may create a little bit of overhead on frontend in my solution. In my solution, it would be probably hard to find and delete audiences based on their id, since I use `Dict parenFolderdId (List Audience)`. I had to go through the whole audience list every time (O(n)) I want to find an audience with a specific id and delete it/change it/place it in different folder. An improvement would be to use `Dict parentId (Dict audienceId Audience)` where I could easily access audiences based on their id in O(1). This may work, however, I did not lead this idea to the end and think about all of the implementation details. I am quite interested, how you solved this problem? @Janiczek 

@janjelinek 
> I also noticed inconsistency in approach

I am sorry for that. Of course, that would change if working on a project with specific coding styles.

> Also I miss some more thinking about edge-cases and impossible states

Yes, I should have thought about that more probably.

> I can see only minimal styling here without any external (S)CSS file 

I know that styles are very important on frontend. And I also know that these "embedded" (not external) styles are insufficient. I think, that in a project with its own coding styles, I would definitely style components according to the codebase's style guides. I just did not want to spend much time writing CSS but with actual coding.

> Also you probably forgot remove [WIP] from PR name

Actually, I did not :D I wanted to first get the answers to the questions I asked and based on the answers I would or would not redo the implementation. But since you mentioned that you usually decide what to display in view functions (as I did in my solution) we can consider, that the PR was basically ready.

I have some working experience with Elm already (half a year of part-time job). However, It was 2 years ago and since then I did not write a single line of Elm. (Haskell yes). Back then, when I was building components in Elm, we used [this](https://atomicdesign.bradfrost.com/chapter-2/) approach and I liked it.

Thank you all for review, praise and feedback. Do you think, I should change/refactor the parts, which you suggested?

Janiczek:
@nitrajka 

> I thought that a single audience can be of a single type (either Shared, Authored or Curated).

Sorry, I might have been projecting our _actual_ Audience + AudienceFolder types into this challenge. (It has evolved from the challenge considerably.) So, let's discard that and backtrack. You can still think of audiences as being of just one type (let's deal only with what's in the FIRST_ and SECOND_STEP, yes).

I would still want to hide empty folders, yes. (In practice that doesn't even happen because we delete empty folders automatically.)

> Of course, the first fetch would always be slow.

That's basically the problem we wanted to solve - there is only ever one request (for entity "Question"), for 6MB of JSON response. We've solved it using lazy tree with zipper. So basically we're doing _some_ work after receiving it, but putting most of the work into unexecuted thunks, and those only get forced if user actually navigates the tree hierarchy. So, simplistically, if user opens a "GWI Core" folder for the first time, work happens to find the children of it. If they open another folder, another chunk of work gets done. If they open "GWI Core" again, the work from before is memorized so no more work is needed.

So, thousands of small 0.1s chunks of works, of which many won't be ever executed, are better for the user experience than one huge 3s batch job at the beginning :)

> Do you think, I should change/refactor the parts, which you suggested?

I think you don't need to refactor it any further. The challenge has served its purpose! :)

PR #22, opened 2020-11-17
Elm Challenge Implementation

omedusyo:
Hi

I am interested in the position. My only experience with Elm is this Elm Challenge, but I am familiar with basic Haskell/OCaml/SML so it was not a problem. I also did some toy React applications in ReasonML (which is basically OCaml). Here's my CV https://raw.githubusercontent.com/omedusyo/resume/master/resume.pdf

0. To compile the app, execute
     `$ elm make src/Main.elm --output=public/main.js`
    and open `public/index.html`. The css is in `public/style.css` and is not included when the app is executed via elm reactor.
1. `Data.Audience` and `Data.AudienceFolder` expose dictionaries
    whose keys are folder ids and whose values are `folders`, `subfolders`, and `subaudience `of the mentioned folder id.
2. If all parsing of JSON goes well, the state of the application is just
   a triple (record) of parent folder, its subfolders and subaudiences.
3. The only permissible action is to change the current parent folder.


omedusyo:
READY

Janiczek:
Hello @omedusyo! Cool that you've done the challenge!

# Re hiring:

To be honest the Elm position in my team was filled recently, but there are other Elm teams in the company. Let me check the current situation there and I'll get back to you :pray:

EDIT: we'll probably start writing the "hiring / not hiring" information into this repo's README as a result :smile:

(A summary of my review of the code itself is somewhere below :point_down:)

src/Data/Audience.elm
```diff
@@ -35,6 +45,64 @@ type alias Audience =
...
+
+maudiences0 : Result D.Error (List Audience)
+maudiences0 = D.decodeString audiencesDecoder audiencesJSON
+
+type alias Subaudiences = Dict Int (List Audience)
+
+-- Creates a key/value map where the keys are folder ids, and values are their list of audiences
+-- TODO: can be eta-reduces (eliminate audiences arg), but why bother? This to me is more readable.
+dictOfAudiences : List Audience -> Subaudiences
+dictOfAudiences audiences = fromListAppendBy .folder audiences
```
Janiczek, line 102: :tada: :100: :+1: I am on the same side of the spectrum :smile:

http://martin.janiczek.cz/clanek/being-clever-antipattern-in-elm/

src/Data/AudienceFolder.elm
```diff
@@ -23,6 +30,44 @@ type alias AudienceFolder =
...
+nameField = D.field "name" D.string
+parentField = D.field "parent" (D.nullable D.int)
+
+-- = AUDIENCE FOLDER DECODER --
+folderDecoder : Decoder AudienceFolder
+folderDecoder =
+  D.succeed AudienceFolder
+    |> D.andMap idField
+    |> D.andMap nameField
+    |> D.andMap parentField
```
Janiczek, line 44: This eliminates my usual question of "what would you choose if you ran out of `map8`? :joy: Obviously given your background you know about applicatives. :+1:

src/Data/AudienceFolder.elm
```diff
@@ -23,6 +30,44 @@ type alias AudienceFolder =
     }
 
 
+-- = BASIC FIELDS DECODERS =
+idField = D.field "id" D.int
+nameField = D.field "name" D.string
+parentField = D.field "parent" (D.nullable D.int)
```
Janiczek, line 36: Not a fan of `elm-format`? :slightly_smiling_face:

src/Data/AudienceFolder.elm
```diff
@@ -23,6 +30,44 @@ type alias AudienceFolder =
...
+  D.succeed AudienceFolder
+    |> D.andMap idField
+    |> D.andMap nameField
+    |> D.andMap parentField
+
+foldersDecoder : Decoder (List AudienceFolder)
+foldersDecoder = D.field "data" (D.list folderDecoder)
+
+maudienceFolders0 : Result D.Error (List AudienceFolder)
+maudienceFolders0 = D.decodeString foldersDecoder audienceFoldersJSON
```
Janiczek, line 50: I'm unsure of what the `0` suffix naming means. I believe I haven't seen that even in Haskell codebases etc. Can you elaborate?

src/Dict/Helpers.elm
```diff
@@ -0,0 +1,37 @@
...
+  in List.foldl action initState xs
+
+-- Given for example a list of records (type a)
+--   xs : List a
+-- each of which has a field id : Maybe Int,
+-- then
+--   fromListAppendBy .id xs
+-- computes a dictionary whose keys are the ids and values are lists of all records with that id.
+-- 
+-- Note that this conversion is linear (assuming that getKey is constant complexity) in the length of the list xs.
```
Janiczek, line 20: > is linear in the length

Meaning linear in complexity, as in `O(n)`? Just checking I understand the comment correctly.

Janiczek: Also, sounds like [`Dict.Extra.groupBy`](https://package.elm-lang.org/packages/elm-community/dict-extra/latest/Dict-Extra#groupBy)

src/Main.elm
```diff
@@ -1,10 +1,132 @@
...
+  , subaudiences : List Audience }
+
+type alias State = Maybe ExplicitFolder
+
+-- a helper function - not worth to put into a separate module
+listOfMaybeList : Maybe (List a) -> List a
+listOfMaybeList mxs =
+  case mxs of
+    Just xs -> xs
+    Nothing -> []
```
Janiczek, line 39: `listOfMaybeList = Maybe.withDefault []`?

src/Main.elm
```diff
@@ -1,10 +1,132 @@
...
+--    and open public/index.html. The css is in public/style.css and is not included when the app is executed via elm reactor.
+-- 1. Data.Audience and Data.AudienceFolder expose dictionaries
+--    whose keys are folder ids and whose values are folders, subfolders, and subaudience of the mentioned folder id.
+-- 2. If all parsing of JSON goes well, the state of the application is just
+--    a triple (record) of parent folder, its subfolders and subaudiences.
+-- 3. The only permissible action is to change the current parent folder.
+
+-- ===STATE=== --
+type alias ExplicitFolder =
+  { parentFolder : AudienceFolder
```
Janiczek, line 28: It seems to me like `parentFolder` could be better named `currentFolder` - is that correct characterization?

src/Main.elm
```diff
@@ -1,10 +1,132 @@
...
+listOfMaybeList : Maybe (List a) -> List a
+listOfMaybeList mxs =
+  case mxs of
+    Just xs -> xs
+    Nothing -> []
+
+stateOfFolderId : Int -> State
+stateOfFolderId folderId =
+  case (folders0, subfolders0, subaudiences0) of
+    (Ok foldersDict, Ok subfoldersDict, Ok subaudiencesDict) ->
```
Janiczek, line 44: Interesting use of a precomputation (I guess if we were receiving the data dynamically in runtime, these precomputed dicts would live in Model as a sort of cache?). Out of curiosity, what approach would you choose if the amount of data would be so big that doing this eagerly for all items would be too slow (app unresponsive for x seconds)?

src/Main.elm
```diff
@@ -1,10 +1,132 @@
...
+initState : State
+initState = stateOfFolderId 357
+-- initState = stateOfFolderId 3111
+-- initState = Nothing
+
+-- ===ACTION===
+type Action = ChangeFolder Int
+
+action : Action -> State -> State
+action (ChangeFolder folderId) _ = stateOfFolderId folderId
```
Janiczek, line 64: Unconventional naming! :)
Basically every Elm app I've seen does this as
```elm
type Msg = ...
update : Msg -> Model -> Model
```

src/Main.elm
```diff
@@ -1,10 +1,132 @@
...
+
+folderIcon = Html.img [ Attr.src "folder.svg" ] []
+editIcon = Html.img [ Attr.src "pencil.svg" ] []
+
+-- TODO: It is a bit unfortunate to recompute lengths every time the state changes,
+--       but in practive for the size of the data that I am given this is not a big deal.
+viewParent : ExplicitFolder -> Html Action
+viewParent state =
+  let parentFolder = state.parentFolder
+      size = List.length state.subfolders + List.length state.subaudiences
```
Janiczek, line 89: You are precomputing all the parts of `ExplicitFolder` in advance anyway... You could say it's just a step towards precomputing the `ExplicitFolder`s outright, and some helper data like the `size` while you're at it... But yeah in practice it's fine to put computation into `view` and start optimizing only after that starts becoming a problem. And even then, optimizations like Html.Lazy and Html.Keyed are usually enough, no need to go for the big guns (caching) for a very long time.

Janiczek:
# Re the challenge itself:

Considering this was your first Elm project, it's quite impressive. There are some "best practice" things and common helper library functions that you'd have to learn on the job but I think that's not a problem at all. I have some questions posted in the comments above; I'm most interested in learning your thoughts on what to do when the amount of data starts being a problem for the eager precomputation.

I've had problems changing folders though. I can't get outside the `Demographics` folder. Is that a bug or am I just missing something?

Other than that, it's pretty solid. You obviously know your way around functional programming, the HTML+CSS also looks fine.

I must say, I'm impressed by the combination of facts that you have category theory / advanced FP knowledge, but (!) that your code is pretty normal/readable/no extra bells and whistles (that's a good thing). Sometimes there's a tendency to use abstractions when you know them even if you don't necessarily need them, but I didn't see that here. Good job! :)

src/Data/AudienceFolder.elm
omedusyo: That's just a personal convention. I usually give a suffix (in math notation it would actually be subscript) to variable names that denote "concrete" (or somehow static, persistant) elements in the current context, e.g. I would write `n0 = 123`, or `verySpecificString1 = "hello, world"`. I use this most often in unit tests. But the parameters of some function `f : Int -> a` wouldn't have such a suffix `f n = ...`. The distinction of "concrete" vs "general" variable names is not formal, it's more what feels right to me. In the real codebase we wouldn't have a static json file, we would get it via HTTP request, so the module `Data/AudienceFolder ` would export some sort of a function and not a concrete lists. So I wouldn't actually have to use something like this.

Also it seems that Elm doesn't allow variable shadowing, so that gave me a bunch of headaches - because of this I was forced to use something like suffixes. But later I managed to refactor the code such that it didn't use shadowing (so maybe the ban is not completely evil :smile:  )

src/Data/AudienceFolder.elm
omedusyo: I didn't know it existed. But it seems that vim installation is pretty easy, so I wouldn't have any problem using it.

src/Dict/Helpers.elm
omedusyo: Damn, `Dict.Extra.groupBy` is exactly what I was looking for. And I have no idea how I could have missed it since it's the first function in the module. I actually found `Dict.Extra.fromListBy`, but then reimplemented it myself - I didn't want to include the whole module because of one function for some reason.

omedusyo: Yes, `O(n)`, the worst case asymptotic complexity.

src/Main.elm
omedusyo: Yes, that would have been better.

src/Main.elm
omedusyo: Hmm, yes, definitely :smile:

src/Main.elm
omedusyo: I think I chose `Action, State` naming because I was using React/Redux some time ago. Also we can think of `action : Action -> State -> State` as `\xs, s -> reduce action s xs : List[Action] -> State -> State`. In math this is usually called Monoid Action `(*) : M -> X -> X` of the Monoid `M` on a type `X`. But I am able to live with `update, Msg, Model` too :smile:

omedusyo:
This is the folder structure
AudienceFolders
the folder that's displayed by default is **Demographics**. It has no parent, and it has no subfolders, so we can't really go anywhere from it. If we were to change the initial model to **New Group**, then we could go to **New Group 2** and from it back to **New Group**. Was I also supposed to create views to display all possible folders? I tested it just by changing the initial model in the code :fearful:

src/Main.elm
omedusyo: Yes, I basically precomputed `folders0, subfolders0, subaudiences0` just by going through all of the json blob.

You mean if the json blob was huge?
In that case, I would need to develop some sort of lazy parsing mechanisms to extract only the data from the json blob that I needed at the moment. But this would depend a lot on how the json blob is structured. For example if I wanted to compute all `subaudiences` of a given folder with `folderId0 : Int`, I would have to depend on some data locality, meaning I would have to rely on something like all audiences with the parent `folderId0` would be in one place in the json blob. If they were scattered all over the place, then this approach would not gain a significant speedup.

I think I would like a more specific question with explicit assumptions about the data to really answer this.

I had to deal with similar concerns in https://omedusyo.github.io/combinatorics/# . Here if I chose `n = 20, m = 10`, and did it in a naive way, I would have to compute an array of length `184756`. I resolved this with laziness via es6 generators.

src/Data/AudienceFolder.elm
Janiczek: Yeah the shadowing ban is a bit annoying but it has helped us find some bugs too! So in the end I think I like it :smile:

src/Main.elm
Janiczek: @omedusyo

> You mean if the json blob was huge?

Yes, but let me clarify: for this thought exercise I'm not worried about the duration of parsing JSON, but the preprocessing that you do afterwards.

> I think I would like a more specific question with explicit assumptions about the data to really answer this.

Fair enough :slightly_smiling_face: In our production codebase we have entities `Question` and `Category` (which are roughly similar in hierarchy to `Audience` and `AudienceFolder`: Categories contain Questions), and the JSON blob can sometimes be around 6 MB in size. If you try figuring out all this `Dict CategoryId (List QuestionId, List CategoryId)` parentage stuff eagerly, the page freezes for a few seconds (at least it did when we last tried this). So we had to do something else. I'm interested in what approach other than "do all the preprocessing needed for `view` when receiving the JSON response in `update`" would you try in that situation.

src/Main.elm
Janiczek: Hey, I learned something here :) I didn't know the `M -> X -> X` thing had a name. `X -> X -> X` would ring a bell (magma, right? just a closed operation over X) but I didn't know about Monoid Actions. Thanks!

Janiczek:
@omedusyo 
> the folder that's displayed by default is Demographics. It has no parent, and it has no subfolders, so we can't really go anywhere from it.

See the challenge description: https://github.com/GlobalWebIndex/challenge-elm/blob/master/FIRST_STEP.md
The intention was to have a root level view too (shown by default):
folders

Janiczek:
@omedusyo We've been pretty satisfied / impressed / yadda yadda yadda :smile: and I've checked with the other team leader that he's still hiring :heavy_check_mark: . Can you please send him an e-mail about the job offer to `dhnyk@globalwebindex.com` and discuss the specifics? You can refer to me and this challenge PR. :slightly_smiling_face:

src/Dict/Helpers.elm
omedusyo: Wait, this is more like O(n * log(n))

src/Main.elm
omedusyo: The freezing problem can be solved as follows:
I could do the heavy computation of collecting the data in a separate thread (via Web Workers - I don't know how convenient this is in Elm). While the thread is doing its thing, the state of the app would be something like `SearchingRelevantData` and the view would reflect this. Since the computation is in a separate thread, the main thread would not be frozen.

Now I'll try to think through the question more carefuly.
**Assumptions**: I have a 6 MB `data : String` - the json blob. Since you are not worried about the parsing, I assume that I've parsed the blob into `questions : List Question`, and `categories : List Category`. I also have a `question_id : QuestionId`.

**Goal**: Compute the view that displays the subquestions and subcategories of `question_id`.

Now you say that if I tried to precompute everything into the dictionary that maps category id's to the list of subquestions and subcategories - which amounts to iterating through `questions` and `categories` independently () - then this would freeze the page for too long. I have a feeling that this is about O(n log(n)) amount of work. Well... more than linear, but probably less than O(n log(n)).

But I only need to compute relevant data to `question_id` - this is only linear. I could memoize this operation so when I search for the same id later, it would be constant time. But this seems strange to me - I feel like this shouldn't make the difference between freezing and not-freezing. If the precomputation had quadratic time, then that would be believable :smile: I feel like the parsing would freeze the page too. So I don't really know how you did it :open_mouth:

omedusyo: Another possibility would be to have the data sorted on the server by the question id's ahead of time so I can get it precomputed.

omedusyo:
Oh, I totally missed the default component requirement :sob: But it would be pretty easy to implement - actually a quick way to implement this without changing pretty much any of the code is to add a "virtual" folder to json data whose id is `null`. A nice **hack** that I wouldn't use in real codebase ofcourse :smile:

Thanks a lot for the review. I'll contact him.

PR #24, opened 2021-07-20
feat(elm-challenge): Solution for the first part of the challenge

vpjapp:
Hello,

Here's my attempt at cracking the tree browser nut. Hope you like it :)

### DIFF
--- a/elm-tooling.json
+++ b/elm-tooling.json
@@ -1,0 +1,11 @@
+{
+    "entrypoints": [
+        "./src/Main.elm"
+    ],
+    "tools": {
+        "elm": "0.19.1",
+        "elm-format": "0.8.5",
+        "elm-json": "0.2.10",
+        "elm-test-rs": "1.0.0"
+    }
+}
--- a/package.json
+++ b/package.json
@@ -1,0 +1,32 @@
+{
+  "name": "challenge-elm",
+  "version": "1.0.0",
+  "description": "Elm challenge",
+  "main": "src/index.ts",
+  "directories": {
+    "test": "tests"
+  },
+  "dependencies": {
+    "@parcel/transformer-elm": "^2.0.0-beta.2",
+    "elm-tooling": "^1.3.0",
+    "node-typescript": "^0.1.3",
+    "parcel": "^2.0.0-beta.2"
+  },
+  "devDependencies": {
+    "elm": "^0.19.1-5"
+  },
+  "scripts": {
+    "start": "parcel src/index.html",
+    "test": "elm-test-rs"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/GlobalWebIndex/challenge-elm.git"
+  },
+  "author": "Ville Jäppinen",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/GlobalWebIndex/challenge-elm/issues"
+  },
+  "homepage": "https://github.com/GlobalWebIndex/challenge-elm#readme"
+}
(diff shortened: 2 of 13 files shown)


vpjapp:
READY

elm-tooling.json
```diff
@@ -0,0 +1,11 @@
...
+    "entrypoints": [
+        "./src/Main.elm"
+    ],
+    "tools": {
+        "elm": "0.19.1",
+        "elm-format": "0.8.5",
+        "elm-json": "0.2.10",
+        "elm-test-rs": "1.0.0"
+    }
+}
```
Janiczek, line 11: Nice to see `elm-tooling` used 🙂

package.json
```diff
@@ -0,0 +1,32 @@
...
+    "elm-tooling": "^1.3.0",
+    "node-typescript": "^0.1.3",
+    "parcel": "^2.0.0-beta.2"
+  },
+  "devDependencies": {
+    "elm": "^0.19.1-5"
+  },
+  "scripts": {
+    "start": "parcel src/index.html",
+    "test": "elm-test-rs"
```
Janiczek, line 20: Interesting that you don't need
```js
    "postinstall": "elm-tooling install"
```
here. We usually do 🤷 😄

tests/AudienceTest.elm
```diff
@@ -0,0 +1,40 @@
...
+                    |> (\res ->
+                            case res of
+                                Result.Ok list ->
+                                    list
+
+                                Result.Err _ ->
+                                    []
+                       )
+                    |> List.length
+                    |> Expect.equal 212
```
Janiczek, line 24: Nice idea to make sure the JSON gets parsed correctly.

src/Main.elm
```diff
@@ -1,10 +1,285 @@
...
+
+
+class : String -> Html.Attribute msg
+class =
+    Html.Attributes.class
+
+
+style : String -> String -> Html.Attribute msg
+style =
+    Html.Attributes.style
```
Janiczek, line 254: These probably could have been just exposed unqualified in the imports?
```elm
import Html.Attributes exposing (class, style)
```

src/Main.elm
```diff
@@ -1,10 +1,285 @@
...
+    let
+        classes =
+            "node"
+                ++ (if current then
+                        " current"
+
+                    else
+                        ""
+                   )
+    in
```
Janiczek, line 273: There is this nice function `Html.Attributes.classList`:
```elm
classList
  [ ("node", True)
  , ("current", current)
  ]
```
that could replace this manual conditional string appending

Janiczek (commented):
Overall pretty nice!

Nice that you used Tree.Zipper and made it do the work as needed (JIT), not upfront (AOT).

As you mentioned, it might not make a difference for this particular dataset, but we have had ~6MB worth of data for some other entity type that needed the same tree approach and we also switched to lazy to accomodate for that.

I left some suggestions in the code :)

Janiczek:
Oh and it's probably a bad form to commit the NPM / Yarn lockfiles (`package-lock.json` in your case). But it doesn't matter much here :)

vpjapp:
Hi,

I meant to comment on these earlier, but kinda forgot 😅.

> Oh and it's probably a bad form to commit the NPM / Yarn lockfiles

How come? Isn't it the whole point of the lock file that each team member
can have the exact same versions of each dependency? You know js-ecosystem,
even a patch update may break everything 🙈

> Interesting that you don't need
> "postinstall": "elm-tooling install"

Good point 😄. I think I actually would need one. I just ran it manually.
(Which definitely is bad form).

Thanks for the other comments too. It's nice to get some feedback 👍

Br.
Ville

ma 26. heinäk. 2021 klo 13.36 Martin Janiczek ***@***.***)
kirjoitti:

PR #25, opened 2021-07-27
Solution to first part

8n8:
READY Hi, my solution to the first part of the challenge. It was fun, thank you.


Janiczek:
Hey @8n8 thanks for the solution!
We'll take a look and review the code - but I just want to check - are you applying for the job or did you do the challenge just for fun? 🙂

8n8:
Hi @Janiczek, I did it for fun mainly. Although I am looking for a part-time job, 2 days a week. I expect you want someone full time though.

Janiczek:
Correct. Ah, nevermind! :)

8n8:
Hi @Janiczek, are you still hiring? I'm now able to work full time.

shell.nix
```diff
@@ -0,0 +1,11 @@
+{ pkgs ? import (fetchTarball "https://github.com/NixOS/nixpkgs/archive/a165aeceda9f9741d15bc2488425daeb06c0707e.tar.gz") {} }:
+
+pkgs.mkShell {
```
Janiczek, line 3: Nice to see somebody use Nix :)

src/Main.elm
```diff
@@ -1,10 +1,209 @@
...
-import Html exposing (Html)
+import Browser
+import Data.Audience as A
+import Data.AudienceFolder as F
+import Dict
+import Html
+import Html.Attributes as Hat
+import Html.Events as Hev
+import Json.Decode as Jd
+import Parser as P
```
Janiczek: Reading this code, it was surprising seeing a `Parser` import - I immediately thought it's `elm/parser` parser. I'd keep the `Decoder` naming for stuff dealing with JSON, to prevent misunderstandings/surprises :)

src/Main.elm
```diff
@@ -1,10 +1,209 @@
...
+that it gives a really elegant and minimal internal state.
+
+-}
+type alias GoodModel =
+    -- The keys are child IDs and the values are parent IDs. Root
+    -- items are not included. If a map lookup fails then the item
+    -- is root, so there are no internal errors caused by failing map
+    -- lookups. The reason for using a dictionary rather than a list
+    -- is to enforce that a child can have only one parent.
+    { parents : Dict.Dict Int Int
```
Janiczek: Do you know of a way to signal the difference between the Ints in code, instead of in comments? What options do we have, and which one you'd like the most?

src/Main.elm
```diff
@@ -1,10 +1,209 @@
...
+    -- used to detect whether something is a folder or an audience.
+    , audiences : Dict.Dict Int A.AudienceType
+
+    -- The IDs and names for all the folders and audiences. This should
+    -- only be mapped over, not used for lookup. Lookup is bad in this
+    -- case because failure will have to be handled, and it shouldn't
+    -- ever be possible to have an ID without a name. The reason for
+    -- using a dictionary instead of a list is to enforce that an
+    -- item can only have one name.
+    , all : Dict.Dict Int String
```
Janiczek: You're mixing audience and audience folder IDs then, which ties into your comment

> It seems likely that the provided IDs are row IDS from the database, and that folders and audiences are kept in different tables, so the IDs may well conflict. A globally unique ID is needed here.

Just a thought: it might be possible to have two dicts, one for audiences and one for audience folders. Then we wouldn't need to care about ID colisions nor have to assign them new IDs.

src/Main.elm
```diff
@@ -1,10 +1,209 @@
...
+    , parent : Parent
```
Janiczek: This naming seems non-ideal to me: it is being used as "where I am currently in the tree". Perhaps something like `currentlyOpen` or `currentLevel` might signal that better?

src/Main.elm
```diff
@@ -1,10 +1,209 @@
...
+        [ Hat.id "container" ]
+    <|
+        List.concat
+            [ goUpView model
+            , foldersView model
+            , audiencesView model
+            ]
+
+
+goUpView : GoodModel -> List (Html.Html Msg)
```
Janiczek: Since this is a list of 0 or 1 items, it could be a Maybe (`if ... then Nothing else Just (Html.button ...)`).

Also since Html has an identity element (value of type `Html msg` that does nothing), namely `Html.Extra.nothing` which is just `Html.text ""`, this function could just return `Html Msg`: `if ... then Html.nothing else Html.button ...`

src/Main.elm
```diff
@@ -1,10 +1,209 @@
...
+            [ Html.text "Go up" ]
+        ]
+
+
+foldersView : GoodModel -> List (Html.Html Msg)
+foldersView model =
+    List.map oneFolderView <|
+        Dict.toList <|
+            Dict.filter (isFolder model.audiences) <|
+                Dict.filter (isChildOf model.parents model.parent) model.all
```
Janiczek: Just a personal code preference: using `<|` causes indentation on each new step. Using `|>` doesn't do that:
```elm
model.all
    |> Dict.filter (isChildOf model.parents model.parent)
    |> Dict.filter (isFolder model.audiences)
    |> List.map oneFolderView
```

src/Main.elm
```diff
@@ -1,10 +1,209 @@
...
+
+
+isChildOf : Dict.Dict Int Int -> Parent -> Int -> String -> Bool
+isChildOf parents parent potentialChild _ =
```
Janiczek: This function has an unused `String` argument just because it's used in `Dict.filter`. It would be cleaner to remove the extra argument and call Dict.filter with `(\id _ -> isChildOf model.parents model.parent id)` IMHO.

src/Main.elm
```diff
@@ -1,10 +1,209 @@
...
+audiencesView model =
+    List.map oneAudienceView <|
+        Dict.toList <|
+            Dict.filter (isAudience model.audiences) <|
+                Dict.filter (isChildOf model.parents model.parent) model.all
+
+
+isAudience : Dict.Dict Int A.AudienceType -> Int -> String -> Bool
+isAudience audiences id _ =
+    not <| Dict.get id audiences == Nothing
```
Janiczek:
```suggestion
    Dict.get id audiences /= Nothing
```

README.md
```diff
@@ -1,3 +1,32 @@
...
+3. Run ```nix-shell --pure``` and wait while Nix installs the dependencies.
+4. Run ```./build.sh``` to run the tests and build.
+
+## Run
+
+Run a local webserver in the ```static``` directory, for example ```python3 -m http.server```.
+
+## UI
+
+I decided not to spend time making it pretty, so there aren't even folder counts, or folder names at the top of opened folders. But these things are there in the model, so would be easy to put in.
```
Janiczek, line 20: When first running the project, I kept looking at whether I was missing some CSS file. A little polish wouldn't hurt, but anyways beauty wasn't the point of the challenge :)

src/Main.elm
Janiczek: Also, a question: is there some alternative way how to model this navigation breadcrumb problem without needing to look up whether a folder has a parent folder?

src/Parser.elm
```diff
@@ -0,0 +1,374 @@
...
+
+
+type alias MakeModel =
+    { folders : List F.AudienceFolder
+    , audiences : List A.Audience
+    , model : Data
+    , unique : Int
+    , folderIds : Dict.Dict Int Int
+    , audienceIds : Dict.Dict Int Int
+    }
```
Janiczek: Seems to me like it would be cleaner to create the fields separately using some List.map or List.indexedMap or List.foldl passes, rather than try to keep track of all the fields in one `addAudience`-like function.

Or at least I have a hard time verifying there is or is not a bug somewhere in them! :) With the separate computations it would be IMHO more localised and easier to check.

src/Parser.elm
```diff
@@ -0,0 +1,374 @@
...
+        newId =
+            m.unique
+
+        ( newParentId, newUnique ) =
+            case Dict.get parentId m.folderIds of
+                Nothing ->
+                    ( m.unique + 1, m.unique + 2 )
+
+                Just exists ->
+                    ( exists, m.unique + 1 )
```
Janiczek: Re the previous comment near `MakeModel`: this for example is something I'd love to _not_ have to write. It could be circumvented using something like
```elm
let
  audiencesCount = List.length audiences
  foldersCount = List.length audienceFolders
  newAudienceIds = List.range 0 (audiencesCount - 1)
  newAudienceFolderIds = List.range audiencesCount (audiencesCount + foldersCount - 1)
in
...
```
and then just zipping them with the audiences list so that you have both the ID and the audience it should be applied to, in the same lambda.

But again, I think a cleaner solution would be to _not_ give new IDs to the audiences and folders.

src/Parser.elm
```diff
@@ -0,0 +1,374 @@
...
+                        String.concat
+                            [ "parent with ID "
+                            , String.fromInt parent
+                            , " and child "
+                            , String.fromInt child
+                            , " does not exist"
+                            ]
+
+                Just _ ->
+                    allParentsExist all arents
```
Janiczek: There are nicer solutions possible than the explicit recursion:
* if you're OK with just a `Bool` output (no further data about what's missing, just that _something's_ missing) then `List.all (parentExist all) parents` would do just fine
* if you want all errors, you can `List.filterMap ... parents`

The current solution returns the first error but not all of them.

src/Parser.elm
```diff
@@ -0,0 +1,374 @@
...
+maybeChain maybes =
+    case maybes of
+        [] ->
+            Nothing
+
+        Nothing :: aybes ->
+            maybeChain aybes
+
+        (Just m) :: _ ->
+            Just m
```
Janiczek: For fun, a possible reimplementation:
```elm
maybeChain maybes =
  maybes
    |> List.filterMap identity
    |> Maybe.andThen List.head
    |> Maybe.map Just
    |> Maybe.withDefault Nothing
```

```elm
maybeChain maybes =
    List.foldl Maybe.Extra.orElse Nothing maybes
```

tests/Tests.elm
```diff
@@ -0,0 +1,124 @@
...
+
+folderDecoder : List Test
+folderDecoder =
+    [ test "simple folder decoder" <|
+        \_ ->
+            Expect.ok <|
+                Jd.decodeString
+                    P.decodeFolders
+                    F.audienceFoldersJSON
+    , fuzz folderFuzz "fuzz encode-decode for folder" <|
```
Janiczek: Kudos for fuzz tests!! :tada:

Janiczek (approved):
The solution works - that's the main thing 🙂 

Sometimes it was a bit overengineered and IMHO there were some decisions early on in the type modelling phase that later led you to way too complicated solutions, but you've dealt with it 👍 

I've left some questions prodding for alternative solutions - if you have the time and are in the mood for it, you can try to answer these and we can discuss them further.

Thanks for the tip for the `Data-Oriented Design` book!

----

> Hi @Janiczek, are you still hiring? I'm now able to work full time.

Apologies for not reviewing the code earlier, and thanks for the ping!

I'll let @juanramoncg continue with the hiring process. AFAIK you can expect some Google Meet calls, and you'll probably be expected to send us your CV. But I'll let Juan handle the rest 🙂 

Thanks for the finished challenge!

src/Main.elm
8n8: I agree it's confusing, and have changed it to DataParser. I like having 'Parser' in the name because I like the way this module catches as many errors in the data as possible up front, i.e. I like the 'Parse not validate' school of thought.

src/Main.elm
8n8: I considered having something like:

```
type ChildId
    = ChildId Int
```

and 

```
type ParentId
    = ParentId Int
```

Or I could have type aliases instead of wrappers. One problem with the wrappers is that they are not comparable types in Elm (I think), so can't be used directly as Dict keys. One way to deal with this would be to make a module containing a Dict Int Int structure, but only export getters and setters for ChildId and ParentId. Something like:

```
module Parent exposing (Parents, ChildId, ParentId, makeChild, makeParent, setParent, getParent)

type Parents =
    Parents_ (Dict Int Int)

type ChildId = ChildId_ Int

type ParentId = ParentId_ Int

setParent : ChildId -> ParentId -> Parents -> Parents
setParent = ...

getParent : ChildId -> Parents -> ParentId
getParent = ...
```

I'm torn on this one. I really like the simplicity of plain Dict Int Int. I think it's also important to bear in mind that indirection has a readability cost, because whenever you see 'ChildId' in the code you have to look up its definition, which is one more bit of cognitive effort. And if it's in a separate module that's even harder. On the other hand, when mapping over the dict, functions will have arguments like ... Int -> Int ... and how can the reader tell which is which? They have to look back to where the function is called. There are the variable names, but they are not checked by the type system.

In general I'm not that keen on type aliases, unless they are for really huge types like record types that are not practical to include in full. They don't improve type safety, and increase mental indirection for the reader.

I am perhaps over-reacting slightly to some Haskell code I have seen that has so many type aliases and newtypes that it's really hard to know what's going on, and what the actual data is underneath. Elm doesn't seem to suffer from this problem so much.

I also feel that in this case it would be hard for the code to be subtly wrong if somewhere the child and parent IDs got mixed up. I think it would be noticeable straight away.

When in doubt, I tend to go with less code is better.

src/Main.elm
8n8: Yes I like this idea. And I just realised that if I assume names are unique within a folder, I don't need audience IDs at all, because an audience is uniquely identified by the combination of its name and its parent ID. So now I only have folder IDs and can use the original ones. In this new Model, I always store folder names alongside the folder IDs, because they always have a name, and this prevents having to look up names, which could fail. I can also drop 135 lines of complicated parser code, so hopefully it won't seem so over-engineered.

src/Main.elm
8n8: OK agreed, have changed 'parent' to 'currentLevel'.

I actually feel pretty happy with doing dict lookups to get the breadcrumb. The only objection I can think of is a performance one. But say there are usually less than 10 levels deep of folders, and now I have split folder parents into their own dict, say that there are usually around 1000ish folders in the system. That's 10 dict lookups in fairly small dicts for each folder click to get a proper breadcrumb like A -> B -> C -> D. This seems all right to me.

src/Main.elm
8n8: Oh that's neat. I had the impression that empty text would still affect the DOM somehow, but I've just done a little test and I see that it doesn't.

src/Main.elm
8n8: Yours looks neater visually, but I find it harder to understand. I think it's all the years of using languages that have the arguments in parenthesis has made me feel that arguments must go on the right. Perhaps I should try to get used to it this way.

src/Main.elm
Janiczek: > One problem with the wrappers is that they are not comparable types in Elm (I think), so can't be used directly as Dict keys.

FYI there are packages like [`turboMaCk/any-dict`](https://package.elm-lang.org/packages/turboMaCk/any-dict/latest/) (more performant, needs you to provide a `toComparable` function) and [`pzp1997/assoc-list`](https://package.elm-lang.org/packages/pzp1997/assoc-list/latest/) (less performant, but nicer library API) with which you can have a `Dict customtype a`

src/Main.elm
8n8: I think I actually prefer my way slightly because at least the hole has an explicit type so is arguably easier to read.

src/Main.elm
8n8: Yes this is better.

README.md
8n8: If I get time I'll have a go at making it look nice. I would also like a proper breadcrumb.

src/Main.elm
Janiczek: IMHO even the type aliases would be better (would convey more meaning) than `Dict Int Int` as it could mean anything.

I would go with the custom type wrappers though.

In our real app code we go even further and do it with phantom types:

```elm
type Id a
    = Id String

type AudienceIdTag
    = AudienceIdTag

type alias AudienceId =
    Id AudienceIdTag

type AudienceFolderIdTag
    = AudienceFolderIdTag

type alias AudienceFolderId =
    Id AudienceFolderIdTag
```
which allows us to have some common logic for `Dict (Id tag) ...` and `Set (Id tag)`.

src/Main.elm
Janiczek: I definitely didn't mean it as "you SHOULD do it this way"! :) it's just that I find it nicer and wanted to share that possibility in case you didn't know it.

src/Main.elm
Janiczek: Yeah I think the Dict approach is OK given the challenge constraints (we don't need to show a breadcrumb trail etc.).
Another solution I could think of is a stack of the IDs, so you would have
```elm
path : List AudienceFolderId
```
where the current parent is at the start and as you go to the right you get grandparent etc... all the way to the root.
That way any manipulations are very quick (adding / removing items from left) and you can eg. easily get the breadcrumbs with List.map and perhaps a List.reverse.

tests/Tests.elm
8n8: Thanks. Fuzzing is great for parsers.

src/Parser.elm
8n8: These are fun. This |> style is growing on me. I think I need to use it more.

src/Parser.elm
8n8: I agree that mine is a bit ugly, but what I quite like is that it quits the map on the first error, which is what I want. Though I suppose it doesn't matter finding all of them; it might even help. In my new refactoring, I made a badMap function, for occasions when you want to quit the map when something goes wrong:

```
badMap : (a -> Result b c) -> List a -> Result b (List c)              
badMap f as_ =                                                         
    badMapHelp f as_ []                                                

                                                                       
badMapHelp : (a -> Result b c) -> List a -> List c -> Result b (List c)
badMapHelp f as_ accum =                                               
    case as_ of
        [] ->                                                          
            Ok (List.reverse accum)                                    

        a :: s ->                                                      
            case f a of
                Err err ->                                             
                    Err err                                            
    
                Ok ok ->                                               
                    badMapHelp f s (ok :: accum) 
```

src/Main.elm
8n8: The more I look at it the more I like it.

src/Parser.elm
8n8: Yes I agree. I have now changed it so that the fields are accumulated separately, and it's also much simpler now it doesn't calculate new IDs, following your suggestion above.

src/Main.elm
8n8: This dict is now gone, following your comments below, but one nice thing about using plain ints is that if I choose the IDs carefully, Dict Int Int could be changed to Array Int. So the child IDs are the array positions and the parent IDs are the array contents. I would think that the array lookups would be way faster than the dict lookups, but it's hard to be sure without benchmarking.

Janiczek: Well, we can always benchmark!
```elm
module ArrayDictAccess exposing (main)

import Array
import Benchmark exposing (..)
import Benchmark.Runner exposing (BenchmarkProgram, program)
import Dict


main : BenchmarkProgram
main =
    program suite


suite : Benchmark
suite =
    let
        sampleDict =
            List.range 0 10000
                |> List.map (\i -> ( i, 3 * i ))
                |> Dict.fromList

        sampleArray =
            List.range 0 10000
                |> List.map (\i -> 3 * i)
                |> Array.fromList
    in
    Benchmark.compare "Access"
        "Dict"
        (\_ -> Dict.get 5000 sampleDict)
        "Array"
        (\_ -> Array.get 5000 sampleArray)
```
Screenshot 2021-09-15 at 12 01 57

8n8: Wow that looks good! Is that BrianHicks/elm-benchmark you're using? It says it's not compatible with 0.19.

Janiczek: @8n8 Correct, it just got moved to [`elm-explorations/benchmark`](https://package.elm-lang.org/packages/elm-explorations/benchmark/latest/)! :)

juanramoncg:
Hi @Janiczek thanks for introduce me and thanks @8n8 for doing the challenge. We have an open Senior FrontEnd position in our squad. If you don't mind I'd like to arrange a call so that I can talk a bit more about this position. Could you send me an email with your CV and availability to juan@gwi.com and then we can continue from there?

src/Main.elm
8n8: I've just refactored it again, and now all the data is in arrays of basic types, with the array positions acting as IDs. There's a slight sacrifice of type safety, because I couldn't think of a way to guarantee in the types that everything has a name. But all the lookups are fast array lookups instead of dict lookups. I also added a breadcrumb.

8n8: I just thought of a way to avoid the loss of type safety, by giving the failing name lookups the meaning that the failing ID is the root folder, which doesn't have a name.

8n8: So now the whole internal state is just:
```
type alias GoodModel =
    { names : Array.Array String
    , parents : Array.Array Int
    , firstAudienceId : Int
    , currentLevel : Int
    }
```

juanramoncg:
Hi @8n8 not sure you got my previous message. Pinging!

8n8:
Hi @juanramoncg, thanks for ping. I did see your previous message. I emailed you my CV on 16th Sep. Perhaps it ended up in Spam? I'll send it again.

PR #26, opened 2021-09-17
READY

azkarim:


src/Main.elm
```diff
@@ -1,10 +1,571 @@
...
+
+List Node is then ordered in a Rose Tree using zwilias/elm-rosetree/1.5.0 library.
+
+At first all nodes are entered into rose tree with flat architecture, see flatTree
+Parent child relation is ignored.
+The reason for this is, you cannot put a node into its parent when the parent itself is not yet traversed and is not in the tree yet!
+Once you have all the nodes in a flat tree and information about { parent : Maybe Node, child : Node }, then 
+1. you can pluck a node from the tree
+2. find its parent in the tree
+3. put it inside the parent, outputting updated tree
```
Janiczek: Here's a magic trick: it is possible to create a tree using this parent-child information. There's a nice function for it: [`Tree.unfold`](https://package.elm-lang.org/packages/zwilias/elm-rosetree/latest/Tree#unfold)

Example usage: https://ellie-app.com/fgPcP2M5wPda1

I think it would remove the need for all these functions:
* `flatTreeToParentChildTree`
* `reorderNode`
* `appendASubTreeAsChild`
* `findASubTree`
* `removeASubTree`
* and probably more

Anyways, it's impressive that you were able to do things explicitly, by hand... :)

src/Main.elm
```diff
@@ -1,10 +1,571 @@
...
+        |> appendASubTreeAsChild subTree parentChild
+
+
+appendASubTreeAsChild : Maybe (Tree Node) -> { parent : Node, child : Node } -> Maybe (Tree Node) -> Maybe (Tree Node)
+appendASubTreeAsChild subTree parentChild tree =
+    let
+        appendTreeToTree tree_ =
+            Tree.appendChild (Maybe.withDefault defaultTree subTree) tree_
+    in
+    Zipper.fromTree (Maybe.withDefault defaultTree tree)
```
Janiczek: You're doing `Zipper.fromTree` and `Zipper.toTree` in many of these functions all over again. It could be nicer to make your functions work on `Zipper` and only convert trees from/to zippers at the very boundaries of these pipelines.

src/Main.elm
```diff
@@ -1,10 +1,571 @@
...
 
-main : Html msg
+type Node
+    = Dir Id AudienceFolder
+    | File Id Audience
+    | Root
+
+
+type alias Id =
+    Int
```
Janiczek: Both `Audience`s and `AudienceFolder`s seem to share the same `Id` space in your implementation, although they have separate `Id` spaces in the database (each has its own table with its own autoincrement sequence).

What alternatives to the above types would you suggest if we wanted to keep the audience and audience folder IDs separate and non-switchable?

src/Main.elm
```diff
@@ -1,10 +1,571 @@
...
+removeASubTree node tree =
+    Zipper.fromTree (Maybe.withDefault defaultTree tree)
+        |> Zipper.findFromRoot (\id -> id == node)
+        |> Maybe.andThen Zipper.removeTree
+        |> Maybe.map Zipper.toTree
+
+
+defaultTree : Tree.Tree Node
+defaultTree =
+    Tree.singleton (File defaultAudience.id defaultAudience)
```
Janiczek: The `defaultTree` seems crucial to have for your representation of the problem (`Maybe (Tree Node)`); I just want to suggest that it might be not needed at all.

I'd rather keep the Maybes until the very end and show an error to the user in case of `Nothing` -- rather than show them a made-up data like 
```elm
 defaultAudienceFolder : AudienceFolder
 defaultAudienceFolder =
     { id = 0
     , name = ""
     , parent = Nothing
     }

 defaultAudience : Audience
 defaultAudience =
     { id = 0
     , name = ""
     , type_ = Curated
     , folder = Nothing
     }
```

Also there seems to be a lot of intermediate `Maybe.withDefault` unwrapping. Again, staying within Zippers inside these helper functions might help alleviate this need for defaulting to made-up data.

src/Main.elm
```diff
@@ -1,10 +1,571 @@
...
+allNodes : List Node
+allNodes =
+    let
+        audienceFolders : List AudienceFolder
+        audienceFolders =
+            Maybe.withDefault [ defaultAudienceFolder ] decodeAudienceFolders
+
+        audiences : List Audience
+        audiences =
+            Maybe.withDefault [ defaultAudience ] decodeAudiences
```
Janiczek: Again I'd rather show an error than made-up data.

src/Main.elm
```diff
@@ -1,10 +1,571 @@
...
+
+
+
+-- VIEW
+
+
+type alias Model =
+    { directory : Tree Node
+    , currentTree : Maybe (Tree Node)
+    , zipperStep : Maybe (Zipper.Zipper Node)
```
Janiczek: What's the reason for keeping multiple representations of the same data?

If you have a Zipper, you have the whole tree + information about where you're currently focused on, so that suggests to me you could have just

```elm
type alias Model =
  Maybe (Zipper Node)
```
(the `Maybe` for the possibility of not being able to decode JSON at the beginning, etc.).

`Zipper.fromTree : Tree a -> Zipper a` and `Zipper.toTree : Zipper a -> Tree a` allow you to convert back and forth without loss of information, if you'd need to use some `Tree.*` functions for example.

src/Main.elm
```diff
@@ -1,10 +1,571 @@
...
+        nodes : List Node
+        nodes =
+            case model.currentTree of
+                Just tree ->
+                    getChildrenNodes tree
+
+                Nothing ->
+                    []
+    in
+    div [ class "w-screen h-screen bg-gray-50" ]
```
Janiczek: Just a random note: I was very impressed by the UI! Nice job!

src/Main.elm
```diff
@@ -1,10 +1,571 @@
...
+                , span [ class "inline-block ml-2 text-white font-bold" ] [ text hasParent ]
+                ]
+
+        False ->
+            text ""
+
+
+type ViewType
+    = ViewFolder
+    | ViewFile
```
Janiczek, line 286: Nice usage of custom types! 👏

src/Main.elm
```diff
@@ -1,10 +1,571 @@
...
+    in
+    case String.length hasParent > 0 of
+        True ->
+            div [ class "w-full bg-gray-800 px-4 py-4 cursor-pointer", onClick GoUp ]
+                [ span [ class "text-white " ] [ FA.icon FA.arrowAltLeft ]
+                , span [ class "inline-block ml-2 text-white font-bold" ] [ text hasParent ]
+                ]
+
+        False ->
+            text ""
```
Janiczek: Random remark: there is a `List.Extra.nothing` helper that does `text ""` under the hood but perhaps tells the story of what's happening better :)

src/Main.elm
```diff
@@ -1,10 +1,571 @@
...
+        name node =
+            case node of
+                Dir id audienceFolder ->
+                    audienceFolder.name
+
+                File id audience ->
+                    audience.name
+
+                Root ->
+                    ""
```
Janiczek: This could be a top-level declaration I think. Seems general enough!

src/Main.elm
```diff
@@ -1,10 +1,571 @@
...
+    in
+    nodesToBeViewed
+        |> List.map (\node -> viewNode (name node) viewType (GoTo node))
```
Janiczek: The `viewNode` function could run `name` itself, it doesn't need to be drip-fed that name pre-prepared :)

src/Main.elm
```diff
@@ -1,10 +1,571 @@
...
+
+
+type alias ViewConfig =
+    { icon : Html Msg
+    , isDark : Bool
+    , cursorPointer : ( String, Bool )
```
Janiczek: What does the `Bool` do? It's always True. Sounds like you wouldn't need to track it; and in the `classList` you could do `[..., (cursorPointer, True)]` instead

src/Main.elm
```diff
@@ -1,10 +1,571 @@
...
+
+                Root ->
+                    ""
+    in
+    nodesToBeViewed
+        |> List.map (\node -> viewNode (name node) viewType (GoTo node))
+
+
+type alias ViewConfig =
+    { icon : Html Msg
```
Janiczek: Since sometimes there is an icon and sometimes there's not, this might be better represented as `icon : Maybe Icon`? The `Html Msg` gives too much freedom: there's nothing stopping me from putting `Html.div [] [Html.textarea [] [...]]` in there.

src/Main.elm
```diff
@@ -1,10 +1,571 @@
...
+            Decode.succeed Curated
+
+        "shared" ->
+            Decode.succeed Shared
+
+        "user" ->
+            Decode.succeed Authored
+
+        _ ->
+            Decode.fail "\"type\" is malformed"
```
Janiczek: Nice. As a bonus, it's sometimes useful to show what the failing value was in `Decode.fail`:

```elm
Decode.fail <| "\"type\" is malformed: " ++ name
```

Janiczek:
Overall I like it! The Tree+Zipper is a great choice, and we're using it in production for large trees with lots and lots of data, and basically anytime we need that "breadcrumb"/"remember your position and navigate through the structure" functionality.

The UI, again, is really nice too. Are you happy with Tailwind CSS?

I've left some questions and suggestions in the code; there are some spots that could be improved, like what you store in Model and how you create the tree+zipper initially.

Again, overall, pretty solid!

Feel free to respond to the comments; I'll let one of our engineering managers take over the hiring conversation in the meantime :)

src/Main.elm
azkarim: Thanks for the example. I have noticed `Tree.unfold` but didn't quite get it at the time.

src/Main.elm
azkarim: How about `type alias AudienceId = Int` and `type alias AudienceFolderId = Int` ? It's still somewhat switchable parameter-wise. We can do `type AudienceId = AudienceId Int` to be more explicit in the code.

src/Main.elm
azkarim: Thanks :)

src/Main.elm
azkarim: Yes absolutely. Anything that documents better :) Btw I couldn't find that function in [List.Extra](https://package.elm-lang.org/packages/elm-community/list-extra/8.5.1/)

src/Main.elm
azkarim: Yeh that's right! It's much more neat now.

src/Main.elm
Janiczek: Oh sorry.. I wrote List.Extra but I meant Html.Extra 🤦

azkarim:
I have made some improvement where you suggested. I'm just stuck with how to make `Tree.unfold` accept `Maybe (Node, List Id)` so I can have `Maybe (Tree Node)` that way it's `Maybe` all the way and unwrapping it at the boundary.

```elm
directoryTree : Maybe (Tree Node)
directoryTree =
    let
        {--
        buildTree : Id -> ( Node, List Id )
        buildTree parentId =
            parentChildNodesOnlyDict
                |> Maybe.andThen (Dict.get parentId)
                |> Maybe.withDefault []
                |> List.map nodeToId
                |> Tuple.pair (idToNode parentId)
                --}
        buildTree : Id -> Maybe ( Node, List Id )
        buildTree parentId =
            parentChildNodesOnlyDict
                |> Maybe.andThen (Dict.get parentId)
                |> Maybe.map (List.map nodeToId)
                |> (\maybeIds -> Maybe.map2 Tuple.pair (idToNode parentId) maybeIds)

    in
    Tree.unfold buildTree 0
```
Also `idToNode` becomes `Maybe Node`
```elm
idToNode : Id -> Maybe Node
idToNode id =
    allNodesInDict
    |> Maybe.andThen (Dict.get id)
```

As for Tailwind CSS, I love it :D

src/Main.elm
azkarim: I settled went with
```elm
type Node
    = Dir AudienceFolderId AudienceFolder
    | File AudienceId Audience
    | Root


type alias Id =
    Int


type AudienceId
    = AudienceId Id


type AudienceFolderId
    = AudienceFolderId Id


type RootId
    = RootId Id


type NodeId
    = NodeDir AudienceFolderId
    | NodeFile AudienceId
    | NodeRoot RootId
```
The reason for additional `NodeId` so I can have a generic function like `nodeIdToNode : NodeId -> Node`

juanramoncg:
Hi @azkarim, it is Juan from Users Management squad. We have an open Senior FrontEnd position in our squad. If you don't mind I'd like to arrange a call so that I can talk a bit more about this position. Could you send me an email with your CV and availability to juan@gwi.com and then we can continue from there?

PR #27, opened 2022-05-15
Challenge

lambduli:
It should be almost ready. I just want to write some comments and thoughts.


lambduli:
So here is couple of my thoughts.

I spent quite some time thinking abut the root folder being a special case of a folder.
From the experience - it seems to be worth it to represent it as such, just give it an empty name and `Nothing` as its `id`.
Upside of that is the simple utilization of the `Tree.Zipper` data structure.

Thought, there is a point to be made about removing the root of the structure out and representing it with a different data type.
Something like:
```elm
type Root = Root { audiences : List Audience, folders : Tree AudienceFolder }
```
Then `Tree` would be something like a `rose tree`. The issue is - this does not seem to be very compatible with the `Zipper` - there is a function for constructing a zipper from the forest, but ths leads to an inconsistency - when I "parent out" of the forest I should travel into a root, but zipper does not know about that, so I get `Nothing`.

So I am thinking about a more specific data structure which is similar to a rose tree, but root is a special case.

In any case - all of this is just to get a little bit more type safety. Non-root folders then can not have empty names and they must have IDs (`Int` not an `Maybe Int`). Also this changes how a specific `view` function handles the rendering.

I am not sure the outcome is worth possibly having to write custom data structure (well maybe there's something already).
But anyways, that was my thought on that.

lambduli:
Regarding the styling and structuring of the `Html Msg` values.

I am not really happy about it. But I think there is no point in spending much more time fidling with it.

In a team on a project I would just adopt the style of the rest of the team by looking at existing code.

lambduli:
`READY`

lambduli:
`READY`

src/Main.elm
```diff
@@ -1,10 +1,177 @@
...
+    let
+        (Info { audiences, name, id }) =
+            Zipper.label zipper
+
+        folders =
+            kids zipper
+
+        {- NOTE: I would not really write it like this in real project. Just thought I would have some fun and perhaps humor someone with a pointfree eye ;) -}
+        ifNotRoot =
+            (flip Maybe.andThen <| id) << always << Just
```
janjelinek, line 61: ❤️ 
I like pointfree, I like it that much, so we established some "Are you fcking nuts with those pointfree?" rule, because I was probably only one who always prefer pointfree. But to be fair, it's valid approach do not use it too much, since it make code more difficult to read for new comers and juniors 🤷🏼

src/Main.elm
```diff
@@ -1,10 +1,177 @@
...
+
+        directories =
+            viewFolders folders
+
+        items =
+            if List.isEmpty audiences then
+                [{- EMPTY FOLDER -}]
+
+            else
+                List.map viewAudience audiences
```
janjelinek, line 88: Is there any particular reason to add extra `if` check here? I believe List.map over empty list has same or very similar performance as  `List.isEmpty []`, so to me it's just additional code noise.

src/Main.elm
```diff
@@ -1,10 +1,177 @@
...
+            Zipper.parent zipper
+
+        {- NOTE: Honestly, I am not happy with the code here. I think some more thought out strategy might help.
+           Like - since we know this function is only called when it IS POSSIBLE to go up, we might leverage that and shift the responsibility
+           of proving that the current level HAS a parent to the caller.
+           But I have spent enought time with this code already so fine tuning it seems like a pointless task - epsecially since I can just write my thoughts here.
+        -}
+        goMessage =
+            Maybe.andThen
+                (Zipper.label >> (\(Info { name }) -> String.Extra.nonEmpty name))
```
janjelinek, line 140: Not sure, if this `nonEmpty` constrain for name is necessary. You should probably trust API in this case. 
Also you are extracting `Zipper.parent zipper` for two cases differently. 
Imagine case where you have parent but has empty name. What will happen?

Button will lead to `parent` but message will be `"go to root"` which will be misleading.

I would do simple case here like 

```Elm
case Zipper.parent zipper of 
    Just parent ->
        viewGoButton ("go to " ++ doExtractingLabelMagicHere) parent
    Nothing ->
        viewGoButton "go to root" (Zipper.root zipper)
        
```        
     
 This has also advantage with not executing `(Zipper.root zipper)` in every case. In you case this will be always executed even if you don't need it.
 
 Also 
 ```Elm
 |> Maybe.map ((++) "go to ")
 |> Maybe.Extra.unwrap "go to root" identity
 ```
 
 is basically 
 
  ```Elm
 |> Maybe.Extra.unwrap "go to root" ((++) "go to ")
 ```

janjelinek (commented):
First thank you for your time with this challenge 👍🏼 
I think you did a good job here, code is little bit "too chatty" for my taste, but overall it looks like you know what are you doing. CSS and styling is probably something you are not enjoying that much, so you simply skipped opportunity to shine here, but its OK, we have good team of designers so only requirement is be able know CSS limits and bring designs to live.

src/Main.elm
lambduli: I must have missed that. I was moving things around a lot in the last hour of working on it. And this is probably the case of cleaning, but not double checking.

src/Main.elm
lambduli: 😄 
Point-Free or Die
- Amar Shah

lambduli: I agree though.

src/Main.elm
lambduli: You are right. This was my major pain point. I knew it just ain't it, but was too tired/stuck. I should have given it some time and see it from a different perspective, but I wanted to submit quickly.
It is pretty obvious that I have needlesly overcomplicated the solution and spiraled into the pit. Should have known better.

Regarding the unnecessary `Zipper.root zipper` - honestly I just didn't even question it - and now it kinda seems funny, because it is obvious that Haskell's laziness made me lazy too.

lambduli: The last point is almost embarrassing honestly.
I should never leave it like this, with identity - which is a dead giveaway anyway.
I probably should have given it an extra day and review my code with a fresh set of eyes.
Well, it is a teachable moment.

janjelinek: No worries, I just had to find some issues so you do not feel to confident 😉

PR #28, opened 2022-07-21
RDY challenge

jaruji:
Hey,

tried my hand at both parts of your Elm challenge. I included some thoughts in the comment at the beginning of Main.elm file, along with the launch instructions - I used the elm-live package instead of elm reactor. As I tried to detach the CSS code (mostly) from the Elm view functions and included it in the index.html file, the CSS in elm reactor will be messed up. Had fun working on this challenge, looking forward to your feedback.  Thanks!

image


src/Data/Audience.elm
```diff
@@ -36,6 +59,56 @@ type alias Audience =
...
+
+decodeDataAudience : Decode.Decoder (List Audience)
+decodeDataAudience =
+    Decode.at [ "data" ] (Decode.list decodeAudience)
+
+
+decodeAudience : Decode.Decoder Audience
+decodeAudience =
+    Decode.succeed Audience
+        |> required "id" Decode.int
```
kutyel, line 73: Bonus points for using the Elm decoder `|>` operator, not many people use this, but I definitely prefer it! 👏🏻 💯

src/Data/Audience.elm
```diff
@@ -36,6 +59,56 @@ type alias Audience =
...
+
+--view
+
+
+view : Audience -> Html msg
+view audience =
+    div
+        [ class "audience"
+        ]
+        [ text (trim audience.name 60)
```
kutyel, line 104: I was wondering: "why is he not using the `String.trim` function?", and then I realised that you are trying to shorten the longer names... not a good idea to do this in the Elm side (IMHO), better to leave the complete name and handle from the CSS part the shortening into the ellipsis (this is what we extensively use in our codebase) 😉

src/Data/AudienceFolder.elm
```diff
@@ -24,6 +36,53 @@ type alias AudienceFolder =
...
+        [ class "folder"
+        , if opened then
+            onClick goUp
+
+          else
+            onClick select
+        ]
+        [ if opened then
+            span [ style "float" "left", style "margin-right" "10px" ]
+                [ Outlined.drive_file_move_rtl 20 (Color <| Color.rgb 96 181 204)
```
kutyel, line 57: This was a bit suspicious behaviour from a UX point of view (IMHO), I would rather see a `Return ◀️` button than the same folder name with a very little icon inside the folder which is almost unnoticeable (very minor though)

src/Helper.elm
```diff
@@ -0,0 +1,8 @@
+module Helper exposing (..)
+
+trim: String -> Int -> String
+trim string len =
```
kutyel, line 4: As mentioned, I don't think this whole file is necessary 👀

src/Main.elm
```diff
@@ -1,10 +1,251 @@
...
+   lists every time the category is changed (and only display values stored at a specific key). Classic dict can only use a comparable key though, so no using AudienceType
+   there. Only way would be to match AudienceType to Int values and use this method to initially create the Dict structure. I think presented solution should be satisfactory,
+   as the performance impact should in this case be hardly noticable and we have a guarantee that the view will always work with accurate data (if we dynamically obtained
+   the data, view would always display it accurately - on the opposite site, the Dict would grow outdated and would therefore need to be updated too, which
+   kind of defeats the purpose of creating it in the first place).
+-}
+{- | Use elm-live to launch the app. Install elm live and launch it with commands:
+
+   npm i -g elm-live
+   elm-live src/Main.elm --open -- --output main.js
```
kutyel, line 34: Thanks for including the instructions from elm-live, I always forget the required arguments and it is true that `elm reactor` falls too short for many things 👍🏻

src/Main.elm
```diff
@@ -1,10 +1,251 @@
 module Main exposing (main)
 
-import Data.Audience
-import Data.AudienceFolder
-import Html exposing (Html)
+{-| Using elm-live for hot reloading to make development just a bit more comfortable. Overall prefer using create-elm-app because of it's debugging features,
```
kutyel, line 3: I don't recommend you `create-elm-app` anymore, it is unmaintained and it does not work on Apple M1 machines, if you google a bit you'll find parcel 2 and vite templates that do exactly the same and are fully customisable 😉

src/Main.elm
```diff
@@ -1,10 +1,251 @@
...
+filterAudienceCategory audiences type_ =
+    {- | Helper function to use with List.filter, separate audiences by type_ -}
+    List.filter
+        (\e ->
+            if e.type_ == type_ then
+                True
+
+            else
+                False
+        )
```
kutyel, line 170:
```suggestion
        (\e ->  e.type_ == type_)
```

src/Main.elm
```diff
@@ -1,10 +1,251 @@
...
+                        True
+
+                    else
+                        case e.parent of
+                            Just parentID ->
+                                if parentID == val then
+                                    True
+
+                                else
+                                    False
```
kutyel, line 191: Again, redundant `if .. else`, you are already returning a boolean!
```suggestion
                                parentID == val
```

src/Main.elm
```diff
@@ -1,10 +1,251 @@
...
+                        True
+
+                    else
+                        case e.folder of
+                            Just folderID ->
+                                if folderID == val then
+                                    True
+
+                                else
+                                    False
```
kutyel, line 224: And again...
```suggestion
                                folderID == val
```

src/Main.elm
```diff
@@ -1,10 +1,251 @@
...
+
+                            _ ->
+                                False
+
+                _ ->
+                    if e.folder == Nothing then
+                        True
+
+                    else
+                        False
```
kutyel, line 234: And again...
```suggestion
                    e.folder == Nothing
```

kutyel (approved):
Overall it looks pretty clean and simple, great work and thanks for your submission! 🙌🏻

src/Main.elm
jaruji: No idea how I didn't realize this sooner, so thank you for opening my eyes 😉 Now I can get rid of all those verbose if elses...

src/Main.elm
jaruji: Didn't know there was an alternative, will definitely take a look!

src/Main.elm
```diff
@@ -1,10 +1,251 @@
...
+                        case model.currentFolderId of
+                            Just id ->
+                                if folder.id == id then
+                                    True
+
+                                else
+                                    False
+
+                            Nothing ->
+                                False
```
janjelinek, line 126: You can simply do
```suggestion
                        Just folder.id == model.currentFolderId
```

src/Main.elm
```diff
@@ -1,10 +1,251 @@
...
+                    filterAudienceFolders model.folders model.currentFolderId
+            )
+            |> div []
+
+        --audiences are handled here
+        , div []
+            [ List.map
+                (\audience ->
+                    Data.Audience.view audience
+                )
```
janjelinek, line 147:
```suggestion
            [ List.map Data.Audience.view
```

src/Main.elm
janjelinek: Even more simplification, only `Just val == e.parent` is enough

janjelinek (commented):
Thx for your time to putting this up. Code looks OK with and with some changes it should be even production ready. 

I didn't find much about your styling skills, but it's partially due the fact, that there is no much room for shining on the other hand what you did looks ok and clean so 👍🏼 

There is lot of you can learn in Elm, but seems like you have some good foundations.

src/Main.elm
jaruji: Now that's a really nice way of writing it 👏. Lesson learned.

jaruji:
Thank you for your feedback so far. Both of you made me realize some minor things I should pay more attention to when writing Elm code, mainly potential simplifications of conditional statements :sweat_smile: . I do believe I have some solid foundations when it comes to Elm as I spent quite some time coding in it, but this is actually the first time someone reviewed my code, so your hints are appreciated!

PR #29, opened 2022-07-26
RDY challenge

left-arm:
Hey, here is only part I of the challenge.

This is pure Elm implementation. No additional css or html files. No additional build tools required.
There are some tests. Github actions and Apple M1 friendly.

One can use `elm reactor` or simply run from command line: `elm make --debug src/Main.elm && open index.html`

Screenshot 2022-07-26 at 15 56 16

### DIFF
--- a/src/Data/Decoder.elm
+++ b/src/Data/Decoder.elm
@@ -1,0 +1,52 @@
+module Data.Decoder exposing (..)
+
+import Data.Audience            exposing (Audience, AudienceType(..))
+import Data.AudienceFolder      exposing (AudienceFolder)
+import Json.Decode as Decode    exposing (Decoder, int, string, nullable, list)
+import Json.Decode.Pipeline     exposing (required, optional)
+
+
+decodeAudienceType : Decoder AudienceType
+decodeAudienceType =
+    let
+        parse s =
+            case s of
+                "user" ->
+                    Decode.succeed Authored
+
+                "curated" ->
+                    Decode.succeed Curated
+
+                "shared" ->
+                    Decode.succeed Shared
+
+                _ ->
+                    Decode.fail <| "Invalid audienceType: " ++ s
+    in
+        string |> Decode.andThen parse
+
+
+decodeAudience : Decoder Audience
+decodeAudience =
+    Decode.succeed Audience
+        |> required "id" int
+        |> required "name" string
+        |> required "type" decodeAudienceType
+        |> optional "folder" (nullable int) Nothing
+
+
+decodeGetAudiences : Decoder (List Audience)
+decodeGetAudiences =
+    Decode.at ["data"] <| list decodeAudience
+
+
+decodeAudienceFolder =
+    Decode.succeed AudienceFolder
+        |> required "id" int
+        |> required "name" string
+        |> optional "parent" (nullable int) Nothing
+
+
+decodeGetAudienceFolders : Decoder (List AudienceFolder)
+decodeGetAudienceFolders =
+    Decode.at ["data"] <| list decodeAudienceFolder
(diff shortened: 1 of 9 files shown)


src/Data/Decoder.elm
```diff
@@ -0,0 +1,52 @@
+module Data.Decoder exposing (..)
+
+import Data.Audience            exposing (Audience, AudienceType(..))
```
kutyel, line 3: mhhh where is elm-format? 😆

tests/DecoderTests.elm
```diff
@@ -0,0 +1,81 @@
...
+
+
+
+audienceDecoderTests : Test
+audienceDecoderTests =
+    describe "Audience API decoder"
+        [ describe "valid audience types"
+              [ test "Authored" <|
+                    \_ ->
+                        Expect.ok <| decodeString decodeAudience """ {
```
kutyel, line 16: Nice addition to add Unit tests, but you could have gone a bit further and test against actually parsed valid values 😉 👍🏻 (instead of just `Expect.ok`)

src/Data/Decoder.elm
janjelinek: Imagine you have to change it for all imports every time you add new long import not fitting to current space :)

src/Main.elm
```diff
@@ -1,10 +1,199 @@
...
+    -- Load audiences and folders concurrently.
+    --
+    -- The real application will execute an appropiate HTTP requests.
+    -- But for testing purposes we will use the provided fixtures.
+    , Cmd.batch
+        [ Task.attempt (Result.withDefault [] >> GotAudiences)
+              <| loadDataTask decodeGetAudiences audiencesJSON
+        , Task.attempt (Result.withDefault [] >> GotAudienceFolders)
+              <| loadDataTask decodeGetAudienceFolders audienceFoldersJSON
+        ]
```
janjelinek, line 59: Above you mentioned your preference for "forward" style and here you are mixing both in one place. 
This would be totally fine and matching your preferences 🤷🏼 
```Elm
loadDataTask decodeGetAudiences audiencesJSON
    |> Task.attempt (Result.withDefault [] >> GotAudiences)
```

src/Main.elm
```diff
@@ -1,10 +1,199 @@
...
+view model =
+    let parent =
+            List.head model.stack
+
+        parentId =
+            case parent of
+                Nothing ->
+                    0
+                Just folder ->
+                    folder.id
```
janjelinek, line 117: This could be potentially danger approach, since `0` is valid integer and what if you will have such a number on BE? In this case we are safe since we own BE service (or BE teams does). In comments earlier you have probably better example with "null" string for `Nothing`.
```Elm
case parentId if
     Nothing ->
         "null"
     Just id ->
         String.fromInt id
```

janjelinek (commented):
First of all thx for your time to doing this challenge.

I was surprised by not having everything formatted by elm-format, but this is not big deal. I like your tests addition and simple approach, but in some places it looks little bit rushed and uneven (some functions has annotations but others don't and I did not saw any patter in your doing).

CSS is basically not existed, so I can see you are not fun of designing things (it's not an issue, we have designers so our job is bring designs to live).

Overall good job 👍🏼

src/Main.elm
left-arm: I agree that this is dangerous and there is a  comment in src/Tree.elm:6 with a remainder to myself with similar code.

src/Main.elm
left-arm: It was said about function composition but I see your point and agree that style somewhat inconsistent :)

left-arm:
> General Acceptance Criteria
> ...
> - `elm/html` must be used for rendering UI.
>

I decided that elm-ui or elm-css are not the option. And I really wanted to keep it small and simple and as a result pure Elm solution without any additional build tools. And you are right about designers - usually I've got a static html & styles from a designer and make it live :)

PR #30, opened 2022-09-26
Challenge READY

Forensor:
Hello, I'm Álvaro 👋🙂.

I've done **Part I** and **II** in separated commits (_Part I implementation_ & _Part II implementation_).

The app is already compiled in the _public_ folder, but you can also run `elm make src/Main.elm --output=public/elm.js` if you want to check how the code transpiles.

It was fun! Let me know about anything else.

Cheers!

### DIFF
--- a/src/Decoder.elm
+++ b/src/Decoder.elm
@@ -1,0 +1,38 @@
+module Decoder exposing (..)
+
+import Data.Audience exposing (Audience, AudienceType(..))
+import Data.AudienceFolder exposing (AudienceFolder)
+import Json.Decode exposing (Decoder, andThen, fail, field, int, map3, map4, maybe, string, succeed)
+
+
+audienceFolderDecoder : Decoder AudienceFolder
+audienceFolderDecoder =
+    map3 AudienceFolder
+        (field "id" int)
+        (field "name" string)
+        (maybe (field "parent" int))
+
+
+audienceTypeDecoder : String -> Decoder AudienceType
+audienceTypeDecoder s =
+    case s of
+        "user" ->
+            succeed Authored
+
+        "shared" ->
+            succeed Shared
+
+        "curated" ->
+            succeed Curated
+
+        _ ->
+            fail <| "AudienceType '" ++ s ++ "' does not exist."
+
+
+audienceDecoder : Decoder Audience
+audienceDecoder =
+    map4 Audience
+        (field "id" int)
+        (field "name" string)
+        (field "type" string |> andThen audienceTypeDecoder)
+        (maybe (field "folder" int))
--- a/src/Main.elm
+++ b/src/Main.elm
@@ -1,10 +1,101 @@
 module Main exposing (main)
 
-import Data.Audience
-import Data.AudienceFolder
-import Html exposing (Html)
+import Browser
+import Data.Audience exposing (Audience, AudienceType(..))
+import Data.AudienceFolder exposing (AudienceFolder)
+import Data.Mocked exposing (DirElem(..), Directory, parsedDirectory)
+import Html exposing (Html, div, text)
+import Html.Attributes exposing (classList)
+import Html.Events exposing (onClick)
+import List exposing (drop, head, length)
+import Maybe.Extra exposing (join)
+import Result exposing (Result(..))
 
 
-main : Html msg
+type alias Model =
+    { directory : Directory
+    , parents : List (Maybe Int)
+    }
+
+
+type Msg
+    = OpenFolder AudienceFolder
+    | GoUp
+
+
+parentIs : Maybe Int -> DirElem -> Bool
+parentIs id de =
+    case de of
+        File audience ->
+            audience.folder == id
+
+        Folder audienceFolder ->
+            audienceFolder.parent == id
+
+
+init : Model
+init =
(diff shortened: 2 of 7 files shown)


src/Main.elm
```diff
@@ -1,10 +1,226 @@
...
+typeIs at de =
+    case ( de, at ) of
+        ( File audience, _ ) ->
+            audience.type_ == at
+
+        ( _, Shared ) ->
+            False
+
+        _ ->
+            True
```
janjelinek, line 48: This is danger because of wildcard and also `True` here looks like bug. Nasty one, because when you add some new type in future, thx wildcard it will not popup in compiler. 
I see that this is meant to be only for File type check, but in that case I would prefer more explanatory name or keep this function next to File type definition (in the same module)

src/Main.elm
```diff
@@ -1,10 +1,226 @@
...
+            [ th []
+                [ filterButtonView Authored model.filterBy
+                ]
+            , th []
+                [ filterButtonView Shared model.filterBy
+                ]
+            , th []
+                [ filterButtonView Curated model.filterBy
+                ]
+            ]
```
janjelinek, line 174: No big deal, but it can be simplified by 
`[Authored, Shared, Curated] |> List map (\t -> 
th []  [ filterButtonView t model.filterBy ] )`

src/Main.elm
```diff
@@ -1,10 +1,226 @@
...
+
+        Folder af ->
+            audienceFolderView af
+
+
+goUpView : Model -> List (Html Msg)
+goUpView model =
+    if length model.parents > 0 then
+        [ div
+            [ onClick GoUp
```
janjelinek, line 210: I prefer usage of specific state in view (like goTo "someId or index") since it prevent user from doing two events (go up twice) by double click as accident. It's again not big deal, but just food for thoughts :)

janjelinek (commented):
First thank you for your time doing our challenge. 
Code looks clean and well thought and everything is working 👍🏼 Some minor things, but nothing major.
Good job.

src/Main.elm
```diff
@@ -1,10 +1,226 @@
...
+                [ filterButtonView Curated model.filterBy
+                ]
+            ]
+        ]
+
+
+audienceView : Audience -> Html Msg
+audienceView a =
+    div
+        [ classList [ ( "dir-elem", True ), ( "file", True ) ]
```
kutyel, line 181: Very minor, but if you are going to have hardcoded values, you can just use [class](https://package.elm-lang.org/packages/elm/html/latest/Html-Attributes#class) 😉 
```suggestion
        [ class "dir-elem", class "file" ]
```

src/Main.elm
```diff
@@ -1,10 +1,226 @@
...
+        ]
+        [ text <| a.name ++ " - [" ++ showAudienceType a.type_ ++ "]"
+        ]
+
+
+audienceFolderView : AudienceFolder -> Html Msg
+audienceFolderView af =
+    div
+        [ onClick <| OpenFolder af
+        , classList [ ( "dir-elem", True ), ( "folder", True ) ]
```
kutyel, line 191: Same here:

```suggestion
        , class "dir-elem folder"
```

src/Decoder.elm
```diff
@@ -0,0 +1,38 @@
...
+import Data.AudienceFolder exposing (AudienceFolder)
+import Json.Decode exposing (Decoder, andThen, fail, field, int, map3, map4, maybe, string, succeed)
+
+
+audienceFolderDecoder : Decoder AudienceFolder
+audienceFolderDecoder =
+    map3 AudienceFolder
+        (field "id" int)
+        (field "name" string)
+        (maybe (field "parent" int))
```
kutyel, line 13: This decoders are perfectly fine, but I'll give you a couple of options to prevent you from relying on `mapN` functions 😉 :

- Use `Decode.andMap` from `json-extra` package, this is the most common way you'll see in all GWI repos.
```elm
audienceFolderDecoder : Decoder AudienceFolder
audienceFolderDecoder =
    Decode.succeed AudienceFolder
        |> Decode.andMap (field "id" int)
        |> Decode.andMap (field "name" string)
        |> Decode.andMap (maybe <| field "parent" int)
```

- Use Richard Feldman's [package](https://package.elm-lang.org/packages/NoRedInk/elm-json-decode-pipeline/latest) for pipeline operator (my personal favourite 🤩)
```elm
audienceFolderDecoder : Decoder AudienceFolder
audienceFolderDecoder =
    Decode.succeed AudienceFolder
        |> required "id" int
        |> required "name" string
        |> optional "parent" int 0
```

kutyel (approved):
Thanks for your time and nice and simple solution! 👍🏻

src/Main.elm
Forensor: Haha, how could I've missed it? 🤦‍♂️

src/Decoder.elm
Forensor: Whoa, it looks much nicer. Didn't know about it. Thanks!

src/Main.elm
Forensor: Oh thanks. It was only to check the type of the `Audience` but it's true that it can be a pain to debug in the future.

Thank you very much for your advices! 😄

PR #31, opened 2022-11-20
Challenge READY - Vít Kostřica

VitKostrica:
Hello,

I've done part 1 - all requirements + added breadcrumbs

I'm using elm-live to run it - `elm-live src/Main.elm -- --debug --output=index.js`

Thanks for your comments in advance.


src/Css/audienceBrowser.css
```diff
@@ -0,0 +1,108 @@
...
+    display: inline;
+    line-height: 10px;
+}
+
+#head {
+    width: 100%;
+    min-height: 40px;
+    border-bottom: 1px solid grey;
+    padding: 5px 5px 10px 5px;
+}
```
janjelinek, line 108: It's good practice to do not use HTML IDs for styling, we have our internal guideline for styling https://github.com/GlobalWebIndex/weak-css

src/Main.elm
```diff
@@ -1,10 +1,197 @@
...
+    initModel
+
+
+initModel : Model
+initModel =
+    { folders = AudFolders.audFolders
+    , opened = { currentId = Nothing, parentId = Nothing, usedIdList = [], parentName = [], state = False }
+    , audience = Aud.audience
+    , breadcrumbs = { breadCrumbName = [], breadCrumbId = [] }
+    }
```
janjelinek, line 60: Since `initModel` is not called in two or more places I would consider move this init logic directly to `init` function.

src/Main.elm
```diff
@@ -1,10 +1,197 @@
...
+    in
+    Html.div [ class "container" ]
+        [ Html.h1 [ class "h1" ]
+            [ Html.text "Audience Browser"
+            ]
+        , Html.div [ class "listContainer" ]
+            [ Html.div [ id "head" ]
+                [ Html.div [ class "breadcrumbsArea" ]
+                    [ Html.a [ class "breadcrumbsText" ] [ text "Home" ]
+                    , Html.ul [ class "breadcrumbsText" ] (List.map (breadCrumbs model.breadcrumbs) (List.reverse model.breadcrumbs.breadCrumbName))
```
janjelinek, line 148: In case like this is better to keep `breadCrumbName` in reverse order inside `Model` so you save this `reverse` operation in view. View part is executed with every update cycle, but manipulation with this list is happening only when you change your path.

src/Main.elm
```diff
@@ -1,10 +1,197 @@
...
+
+    else if list.parent == Nothing && opened.state == False then
+        folderButton
+
+    else
+        text ""
+
+breadCrumbs : Breadcrumbs -> String -> Html msg
+breadCrumbs breadcrumbs list =
+    li [ class "breadcrumbs" ] [ text (" -> " ++ list) ]
```
janjelinek, line 181: I know it's not forced in pro-next code and it's controversial topic, but I prefer to keep function definitions before first usage. Usually when you read code it's from top to bottom and when you have function defined first you already know what is this function about later when it's called. But this rule is somehow not relevant for big files as we have in Elm or in case you have tens of function defined before it's used, because you can't hold everything in your head. In cases like that it's important to have good naming. Here I would consider `breadcrumbView` so it's clear purpose of this function.

src/Main.elm
```diff
@@ -1,10 +1,197 @@
...
+    , usedIdList : List Int
+    , parentName : List String
+    , state : Bool
+    }
+
+
+type alias Breadcrumbs =
+    { breadCrumbName : List String
+    , breadCrumbId : List Int
+    }
```
janjelinek, line 41: This can go easily out of sync. You can have list of names not matching list of Ids. 
Better type would be `List { breadcrumbName: String, breadcrumbId: Int }`.

src/Main.elm
```diff
@@ -1,10 +1,197 @@
...
+    , breadcrumbs : Breadcrumbs
+    }
+
+
+type alias Opened =
+    { currentId : Maybe Int
+    , parentId : Maybe Int
+    , usedIdList : List Int
+    , parentName : List String
+    , state : Bool
```
janjelinek, line 34: This is unnecessary. You already have this information in `currentId` It's either `Just a` so `state = True` or `Nothing`. This breaks rule "single source of truth"

src/Main.elm
```diff
@@ -1,10 +1,197 @@
...
+            Html.button [ class "audience" ] [ text listAudience.name ]
+    in
+    if opened.currentId == listAudience.folder && opened.state == True then
+        audienceButton
+
+    else if listAudience.folder == Nothing && opened.state == False then
+        audienceButton
+
+    else
+        text ""
```
janjelinek, line 197: If you get rid of `state` you can simply do this, because of `folder: Maybe Int`.
```suggestion
    if opened.currentId == listAudience.folder then
        audienceButton

    else
        text ""
```

src/Main.elm
```diff
@@ -1,10 +1,197 @@
...
+            Html.button [ class "folder", onClick (MsgFolderOpened list.id list.name) ] [ text list.name ]
+    in
+    if opened.currentId == list.parent && opened.state == True then
+        folderButton
+
+    else if list.parent == Nothing && opened.state == False then
+        folderButton
+
+    else
+        text ""
```
janjelinek, line 177:
```suggestion
    if opened.currentId == list.parent  then
        folderButton

    else
        text ""
```

src/Main.elm
```diff
@@ -1,10 +1,197 @@
...
+            , Html.ul [ class "list" ]
+                (List.map (openFolder model.opened) model.folders)
+            , Html.ul [ class "list" ]
+                (List.map (viewAudience model.opened) model.audience)
+            ]
+        ]
+
+
+openFolder : Opened -> AudFolders.AudienceFolder -> Html Msg
+openFolder opened list =
```
janjelinek, line 165: Careful with naming `list` in this case is `audienceFolder` same below with `audience`. I smell some `copy/paste` here :)

src/Css/audienceBrowser.css
VitKostrica: ok noted, thanks

src/Main.elm
VitKostrica: I need to get used to naming, still new to me :)

VitKostrica:
Thanks for the comments Jan, I will write it down to not repeat these issues in the future.

src/Data/Audience.elm
```diff
@@ -35,22 +22,43 @@ type alias Audience =
     }
 
 
+audience : List Audience
+audience =
```
Forensor, line 26: Just a little thing, but as you've done with `audFolders`, it's usually nice to have `List`/`Array` value names written in plural for readability.
```suggestion
audiences =
```

src/Main.elm
```diff
@@ -1,10 +1,197 @@
...
+            newOpened.parentId
+
+        newParentName name =
+            name :: newOpened.parentName
+
+        newUsedIdList id =
+            id :: newOpened.usedIdList
+
+        newBreadcrumbpName name =
+            name :: newBreadcrumb.breadCrumbName
```
Forensor, line 88: Although the cons operator (`::`) shows that `Breadcrumbs.breadCrumbName` is a `List`, it could be confusing in a larger codebase. `newBreadcrumb.breadCrumbName` alone can lead to think that it is a single value, and, although we have the typesystem to save us from this sort of things, it is always convinient to make it as easy-viewing as possible for you in the future.

PR #32, opened 2023-08-02
Challenge-elm Ready

Saitorfer:
READY

Hello!
I use http server (npm i -g http-server).
To run the code, please use the commands:
1) http-server -p 8000  ./server/Json
2) npm run dev 

Thank you for your time.


Forensor (commented):
Not actually specified in the challenge but for better UX/UI design I think it'd be cool to have some sort of visual way to know which filter is selected:
image

package.json
```diff
@@ -0,0 +1,14 @@
...
+  "main": "index.js",
+  "scripts": {
+    "dev": "elm-live src/Main.elm -- --debug --output=index.js",
+    "compile": "elm make src/Main.elm --output=index.js",
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC"
+}
```
Forensor, line 14: `http-server` and `elm-live` could be added to `devDependencies` so you could modify the starting script to be just one command for better DX, but it's okay as it is.

server/Json/Audience.Json
Forensor, line 1: Most files get automatically recognised by the browser without the need of an extension, however it feels weird to me setting the extension as `.Json` instead of the standard `.json`. It's also the same with the folder name as `Json` instead of `json`, following the style of the rest of the folders. **Very nitpicking but this is important for case sensitive stuff.**

server/Json/AudienceFolder.Json
Forensor, line 1: Same comment as in `server/Json/Audience.Json`.

Saitorfer:
> Not actually specified in the challenge but for better UX/UI design I think it'd be cool to have some sort of visual way to know which filter is selected: image

I'll keep that in mind, thanks for the advice !

src/Decoder.elm
Forensor, line 1: Decoders look really clean with [`Json.Decode.Pipeline`](https://package.elm-lang.org/packages/NoRedInk/elm-json-decode-pipeline/latest/Json.Decode.Pipeline) 👍🏻

src/Main.elm
```diff
@@ -1,10 +1,362 @@
...
+    , errorMessage : Maybe String
+    , currentFolderId : Maybe Int
+    , selectedCategory : Audience.AudienceType
+    }
+
+
+main : Program () Model Msg
 main =
-    Html.text "There will be app soon!"
+    Browser.element
```
Forensor, line 34: Very nitpicking but [`Browser.element`](https://package.elm-lang.org/packages/elm/browser/latest/Browser#element) is actually for HTML elements managed by Elm that you can embed into any JavaScript project. [`Browser.document`](https://package.elm-lang.org/packages/elm/browser/latest/Browser#document) would be more correct, since yours is a whole application done in Elm. But **[`Browser.element`](https://package.elm-lang.org/packages/elm/browser/latest/Browser#element) gets it done so it's okay.**

src/Main.elm
```diff
@@ -1,10 +1,362 @@
...
+        , expect = Http.expectJson MsgGotResultsAudienceFolder Decoder.decodeAudienceFolders
+        }
+
+
+cmdSearchAudience : Model -> Cmd Msg
+cmdSearchAudience model =
+    Http.get
+        { url = model.urlAudience
+        , expect = Http.expectJson MsgGotResultsAudience Decoder.decodeAudiences
+        }
```
Forensor, line 362: I'd better rename this to both `cmdFetchFolders` and `cmdFetchAudiences` since that's what your functions seem to do. I don't think _Search_ is a correct term here. But extra points for using HTTP calls in the app 👍🏻

src/Main.elm
```diff
@@ -1,10 +1,362 @@
...
+import Http exposing (Error(..))
 
 
-main : Html msg
+type Msg
+    = MsgGotResultsAudienceFolder (Result Http.Error (List AudienceFolder.AudienceFolder))
+    | MsgGotResultsAudience (Result Http.Error (List Audience.Audience))
+    | OpenFolder Int
+    | GoBack
+    | SelectCategory Audience.AudienceType
```
Forensor, line 18: This `Msg` is fired to handle the filters, but they are done by `AudienceType`s, not by _categories_ (which are not used here).
```suggestion
    | SelectAudienceType Audience.AudienceType
```

src/Main.elm
```diff
@@ -1,10 +1,362 @@
...
+
+
+type alias Model =
+    { audienceFolder : List AudienceFolder.AudienceFolder
+    , audience : List Audience.Audience
+    , urlAudienceFolder : String
+    , urlAudience : String
+    , errorMessage : Maybe String
+    , currentFolderId : Maybe Int
+    , selectedCategory : Audience.AudienceType
```
Forensor, line 28: Following the comment above:
```suggestion
    , selectedAudienceType : Audience.AudienceType
```

src/Main.elm
```diff
@@ -1,10 +1,362 @@
...
+
+initModel : Model
+initModel =
+    { audienceFolder = []
+    , audience = []
+    , urlAudienceFolder = "http://localhost:8000/server/Json/AudienceFolder.Json"
+    , urlAudience = "http://localhost:8000/server/Json/Audience.Json"
+    , errorMessage = Nothing
+    , currentFolderId = Nothing
+    , selectedCategory = Audience.Authored
```
Forensor, line 341:
```suggestion
    , selectedAudienceType = Audience.Authored
```

src/Main.elm
```diff
@@ -1,10 +1,362 @@
...
+
+
+categoryButtonsView : List (Html Msg)
+categoryButtonsView =
+    [ div []
+        [ button [ class "category-button", onClick (SelectCategory Audience.Authored) ] [ text "Authored" ]
+        , button [ class "category-button", onClick (SelectCategory Audience.Shared) ] [ text "Shared" ]
+        , button [ class "category-button", onClick (SelectCategory Audience.Curated) ] [ text "Curated" ]
+        ]
+    ]
```
Forensor, line 201: Very nitpicking, but you can make it highly scalable with a `List.map`:
```suggestion
    let
        viewAudienceTypeButton : { textShown : String, audienceType : AudienceType } -> Html Msg
        viewAudienceTypeButton { textShown, audienceType } =
            button [ class "audience-type-button", onClick (SelectAudienceType audienceType) ] [ text textShown ]
    in
    [ div []
        List.map viewAudienceTypeButton [{textShown = "Authored", audienceType = Authored}, ... ]
    ]
```

src/Main.elm
```diff
@@ -1,10 +1,362 @@
...
+            reason
+
+
+
+--Get the parent of the folder, if there is no parent he put the currentFolderId to Nothing so he come back
+--to the inicial screen
+
+
+getActualParent : Int -> Model -> Maybe Int
+getActualParent folderId model =
```
Forensor, line 296: [False-friend](https://dictionary.cambridge.org/dictionary/english/false-friend):
```suggestion
getCurrentParent : Int -> Model -> Maybe Int
getCurrentParent folderId model =
```

src/Main.elm
```diff
@@ -1,10 +1,362 @@
...
+
+
+--VIEW OF THE PARTS
+
+
+viewFilteredAudience : (Audience.Audience -> Bool) -> List Audience.Audience -> List (Html Msg)
+viewFilteredAudience filterFn audienceList =
+    [ ul [ class "audience-container" ]
+        (List.filter filterFn audienceList
+            |> List.map viewComponentAudience
```
Forensor, line 114: Good use of function composition 👍🏻

src/Main.elm
```diff
@@ -1,10 +1,362 @@
...
+import Html.Events exposing (onClick)
+import Http exposing (Error(..))
 
 
-main : Html msg
+type Msg
+    = MsgGotResultsAudienceFolder (Result Http.Error (List AudienceFolder.AudienceFolder))
+    | MsgGotResultsAudience (Result Http.Error (List Audience.Audience))
+    | OpenFolder Int
+    | GoBack
```
Forensor, line 17: Usually it's best to pass some sort of data to avoid `Msg` duplication (the DOM is your enemy). So for example here I'd use something like `GoToParentFolder (Maybe Int)` for this purpose, but it's not a big deal.

Forensor (approved):
Very nitpicking things here and there, but overall **everything looks clean and it works**. Some notable things:

- Good use of function composition
- `selectedCategory : AudienceType` is confusing (categories are not used from the JSON), but it doesn't affect functionality, it's only food for thoughts
- Extra points for getting the JSONs from a server
- Some small misspellings, but nothing to worry about

Good job David, and thanks for your time 👍🏻 .

package.json
Saitorfer: oh, it's true, thank you 😄

src/Main.elm
```diff
@@ -1,10 +1,362 @@
 module Main exposing (main)
```
Beny406, line 1: Hi, thx for finishing the challenge. Its a good start. :+1:  Below you can find hints (some rather advanced) of how code could be improved.

* organize file around business logic / data structure -eg Auidence.elm should contain decoders and Cmd related to it https://youtu.be/XpDsk374LDE?t=443
* using elm-css https://package.elm-lang.org/packages/rtfeldman/elm-css/latest/Css instead of separate css file
* its prefered to have naming of Msgs in form of SomethingHappend not DoSomething https://www.youtube.com/watch?v=w6OVDBqergc&ab_channel=ElmintheSpring
*  instead of having errorMessage field - use RemoteData https://package.elm-lang.org/packages/krisajenkins/remotedata/latest/
*  view
   * functions should not accept the whole model, just parts which are necessary for the function to work => easier to debug in future
   * could be organized/simplified
       * no need to split into such a small functions (its not javascript :))
       * no need to have separate function for different situations, its enough to have one view only with filters changing based on situation. Have a look how condensed the view can be
```elm
view : Model -> Html Msg
view model =
    div [ class "table" ]
        (List.concat
            [ [ div []
                    [ button [ class "category-button", onClick (SelectCategory Audience.Authored) ] [ text "Authored" ]
                    , button [ class "category-button", onClick (SelectCategory Audience.Shared) ] [ text "Shared" ]
                    , button [ class "category-button", onClick (SelectCategory Audience.Curated) ] [ text "Curated" ]
                    ]
              ]
            , if model.currentFolderId /= Nothing then
                [ div
                    [ class "goBack-button", onClick GoBack ]
                    [ text "Go Up" ]
                ]

              else
                []
            , [ div []
                    ((model.audienceFolder
                        |> List.filter (\folder -> folder.parent == model.currentFolderId)
                        |> List.filter (\_ -> model.selectedCategory /= Audience.Shared)
                        |> List.map
                            (\folder ->
                                button
                                    [ class "folder-button"
                                    , onClick (OpenFolder folder.id)
                                    ]
                                    [ text folder.name ]
                            )
                     )
                        ++ (model.audience
                                |> List.filter (\audience -> audience.folder == model.currentFolderId && model.selectedCategory == audience.type_)
                                |> List.map
                                    (\audience ->
                                        button
                                            [ class "audience-button"
                                            ]
                                            [ text audience.name ]
                                    )
                           )
                    )
              ]
            ]
        )
```

Saitorfer: Hello, thank you for your advices, I will bear in mind these tips 😄

